PR #3, opened 2015-02-24
Adjust Typed Racket parts of DrRacket

takikawa:
This PR does three things. First, it updates `insert-large-letters.rkt` to use the up-to-date GUI types that are now part of Typed Racket rather than the old incomplete library. It also disables typechecking for `auto-language.rkt` because it doesn't yet work in modern TR.

Finally, it ports `tooltip.rkt` to Typed Racket.

Note: these commits won't work as-is on TR HEAD because they depend on some upcoming TR changes. But I wanted to get this reviewed ahead of time so that I can merge it immediately after merging the TR fixes. (because the TR fixes would also break the existing `insert-large-letters.rkt` for example)

@rfindler does this seem reasonable?


rfindler:
Looks like a big improvement to me!

Robby

PR #5, opened 2015-05-20
Some fixes to the module browser.

samth:


samth:
@rfindler Any thoughts on this?

rfindler:
The first commit seems obviously good.

The second seems good to, but did you try it out with some program and it made things better?

If so, please do push (rebased) commits! (Or ask me and I can do it.)

Thanks,
Robby

samth:
For the second commit, the first change is needed almost always, I think, and the second change is needed for this program (without the change it doesn't see that it's a `'lib` require).

```
#lang typed/racket

(require (submod math/private/base/base-functions #%type-decl))
``

rfindler:
Thanks.

PR #10, opened 2015-06-28
convert rectangle-intersect.rkt to typed racket

AlexKnauth:


rfindler:
Sorry I keep meaning to look into this but not finding time. The conversion is obviously correct, but since this file is pretty obviously right (at least at the level of guarantees that types provide(!)), I want to make sure that the performance is still okay. In particular, is the rectangle intersection called during event callbacks that are already slow (on the border or over the 16 msec mark)? And if so, what does the contract boundary crossing do to them?

We'd need to sort that out before taking this PR.

AlexKnauth:
Ok. This was just a part of my attempt to convert `blueboxes-gui.rkt` to typed racket, so it's fine if this is closed until I can get that going.

samth:
We should actually measure these costs. These are about the cheapest contracts that TR will generate, so I would hope it's not too expensive.

rfindler:
I agree that measurement is a prereq here but it also seems like a place
that's low on the "worth bothering" list.

Issue #19, opened 2015-11-04
Debugger in ASL doesn't (always) show PITCH


rfindler:
This program, in the language-dialog version of ASL, shows too many frames when you repeatedly click "step" in the debugger (the calls to "f" stack up in the "stack" pane of the debugger):

```
(define (f x) (if (zero? x) x (f x)))
(f 1)
```

This program, in contrast:

```
(define (f x) (f x))
(f 1)
```

does not have that problem.

florence:
This problem can be reproduced in full racket:

1. Turn on "Syntactic Test Suite Coverage"
2. Use the program:

```
#lang racket

(define (loop)
  (begin0 ;; <- for some reason ASL sometimes inserts these
    (loop)))

(loop)
```

3. Start the debugger, and step through. Watch the stack grow!

florence:
I should note: If you don't use the debugger and just run the program tail calls happen property (evidence: the program never runs out of memory). So it seems this has something to do with how the test coverage (errortrace), the debugger, and `begin0` interact.

greghendershott:
Although I barely understand this code, maybe this is because `begin0` needs similar `is-tail?` treatment as `begin`? (Although I guess for the first body not the last?)

https://github.com/racket/drracket/blob/master/drracket/gui-debugger/annotator.rkt#L345-L357

```
            [(begin . bodies)
             (letrec ([traverse
                       (lambda (lst)
                         (if (and (pair? lst) (equal? '() (cdr lst)))
                             `(,(annotate (car lst) bound-vars is-tail? module-name ))
                             (cons (annotate (car lst) bound-vars #f module-name )
                                   (traverse (cdr lst)))))])
               (quasisyntax/loc expr (begin #,@(traverse (syntax->list #'bodies)))))]
            
            [(begin0 . bodies)
             (quasisyntax/loc expr (begin0 #,@(map (lambda (expr)
                                                     (annotate expr bound-vars #f module-name ))
                                                   (syntax->list #'bodies))))]
```

rfindler:
It appears to only be the debugger's annotation. Turning off errortrace in the language dialog and then clicking "debug" and "go" gives the wrong printouts for this program (showing growth):

```
#lang racket
(define (f x)
  (cond
    [(zero? x) x]
    [else
     (with-continuation-mark 'x x
       (begin0
        (begin
          (printf "~s\n" (continuation-mark-set->list
(current-continuation-marks) 'x))
          (f (- x 1)))))]))
(f 10)
[...]

PR #20, opened 2015-11-05
Fix debugger's interaction with `begin0` (Issue #19)

florence:
This commit should be merged into the release branch


florence:
I didn't write any tests since none exist and this is a small change. I can conjure up a testing harness if it seems very important. (But that might not get done before the release)

stamourv:
The change looks reasonable.

florence:
@rfindler Pointed out that we already had a test case for this, so I added it.

rfindler:
I pushed this.

https://github.com/racket/drracket/commit/335e43e07d044b7c060597fdbe099588ea4752dd

PR #21, opened 2015-11-10
added tests for letrec undefined bug

florence:
I've checked, and the new tests fail without e20b40f81a


rfindler:
Thanks! I've pushed the commit.

PR #51, opened 2016-04-15
Fix typo

leafac:
"... when for ..." -> "... for ...".


rfindler:
I would have removed the "for", actually. No?

leafac:
@rfindler, I think either is fine. But I changed according to your suggestion.

PR #53, opened 2016-04-18
typo correction line 45 exports to export

spdegabrielle:
line 45 changed 'they exports all of the bindings of' to 'they export all of the bindings of'


rfindler:
Thanks!

PR #63, opened 2016-06-22
Fix button order

takikawa:
Based on old screenshots like https://en.wikipedia.org/wiki/Racket_%28programming_language%29#/media/File:Drracket.png I think the "debug" button is supposed to come after "check syntax". This PR fixes this ordering for typical uses (switching the language to "R5RS" for example seems to mess up the order, but I think it's an unrelated problem).

### DIFF
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -2069,15 +2069,7 @@
               [n (in-list numbers/fs)]) 
           (add-to-toolbar-buttons 'register-toolbar-buttons b n))
         (set-toolbar-label-visibilities/check-registered)
-        
-        ;; sort panel contents
-        (define panels '())
-        (for ([tb (in-list bs)])
-          (define parent (send tb get-parent))
-          (unless (memq parent panels)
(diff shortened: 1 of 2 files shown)


drracket/drracket/private/unit.rkt
```diff
@@ -2103,6 +2095,14 @@
...
       (define/public (sort-toolbar-buttons-panel)
         (define bp (get-button-panel))
+        (define (cmp panel)
+          (cond
+            [(is-a? panel vertical-pane%) >=]
+            [(is-a? panel horizontal-pane%) <=]
+            [else
+             (if (send panel get-orientation) ;; horizontal is #t
+                 <=
+                 >=)]))
```
rfindler: This cond isn't right. You want just the third case, I believe.

rfindler:
In unit.rkt, did you change the behavior or did you just move stuff around?

rfindler:
There appears to have been a bugfix wrt to a use of `min` and possibly adding a missing `#f` in a `hash-ref`? 

Anyway, this looks good to me. I'll merge the change.

Thanks!

Issue #74, opened 2016-09-17
"could not find "Modern" style" error when compiling documentation for DrRacket on a non-English system


jsmaniac:
When running the scribble file `drracket/scribblings/drracket/interface-essentials.scrbl`, the following error is raised when the system's language is not English (e.g. "Modern" in French is "Moderne" with an "e" at the end, so it's not finding the style).

> could not find "Modern" style

I'll submit a PR in a minute

rfindler:
I think it should just use the string-contant, with no conditional and no dynamic.

jsmaniac:
@rfindler The "Modern" style name is not a string constant, it's a symbol (which is translated by the GUI using `(dynamic-string-constant the-symbol)`).

rfindler:
When DrRacket is in the french mode, this expression: `(string-constant modern-color-scheme)` evaluates to `"Moderne"` and when in English mode it evaluates to `"Modern"`. So using this code:

```
     (filter
      (λ (x) (equal? (hash-ref x 'name)
                     (string-constant modern-color-scheme)))
```

does what you want it to do and does not require `dynamic-string-constant` or the test.

jsmaniac:
@rfindler Oh okay, I misunderstood you. I used `dynamic-string-constant` because that's what the `(let loop ([datum datum]) …` uses below (see [here](https://github.com/racket/drracket/pull/75/files#diff-891fdbb355c0801dd69853f956798706R809)). Should I change it in the loop too?

Also, I used the condition in case, for whatever reason, the scheme `'modern-color-scheme` or the translation for its name was unavailable. I'll remove it, unless you think that might happen in some cases (it would still throw an error while compiling the document anyway)

jsmaniac:
@rfindler Scratch that, I realised string-constant is a macro, and dynamic-string-constant is the corresponding function. Sending an update in a minute.

rfindler:
I don't think the one below should change. It is there for a different
reason.

jsmaniac:
@rfindler I pushed the updated version. Thanks for the review :) !

PR #75, opened 2016-09-17
Fix "could not find "Modern" style" error when compiling the documentation

jsmaniac:
The `filter` operates on a list where the style names are translated, but looked for an untranslated name, giving an error when compiling on a system with a French locale, for example ("Moderne" with an "e" instead of "Modern")


jsmaniac:
This fixes issue #74

rfindler:
Thank you for finding and fixing this!

PR #139, opened 2017-08-25
Move syntax-check arrow colors into color scheme.

shhyou:
This allows the color scheme package to configure the color of the arrows.


shhyou:
Added color setting to TOL's color schemes.

rfindler:
Looks great, thanks!

PR #144, opened 2017-09-20
Display tooltips that are too large, and update docs on positions

### LINKED ISSUE #113: Example of tooltip in documentation doesn’t work
[This section](https://docs.racket-lang.org/tools/Check_Syntax.html?q=check%20syntax#%28part._.Syntax_.Properties_that_.Check_.Syntax_.Looks_.For%29) of the documentation includes the following example:

```racket
#lang racket
(define-syntax (char-span stx)
  (syntax-case stx ()
    [(_ a)
     (syntax-property
      #'a
      'mouse-over-tooltips
      (vector
       stx
       (syntax-position stx)
       (+ (syntax-position stx)
          (syntax-span stx))
       (format "this expression\nspans ~a chars"
               (syntax-span stx))))]))
 
(char-span (+ 1 2))
```

Supposedly, DrRacket should add a tooltip, but it doesn’t:

bug-report

david-christiansen:
Previously, the example code in the documentation for tooltips did not
work unless there was additional whitespace at the end of the
module. This occurred for two reasons: tooltips outside the valid
range for editor positions were ignored, and tooltips are given in
editor positions, which differ from syntax object positions, so the
tooltip stuck out past the end of the file and was ignored.

This commit does two things to fix this:
 1. Tooltip regions are truncated, rather than ignored, when they
    exceed the file length.
 2. The docs mention the difference in indexing, and the sample code
    contains the necessary sub1 calls.

This solution is backwards-compatible - some non-working programs now
work, but working programs won't change the semantics of tooltip
placement.

Fixes #113.

### DIFF
--- a/drracket/drracket/private/syncheck/gui.rkt
+++ b/drracket/drracket/private/syncheck/gui.rkt
@@ -1052,8 +1052,13 @@ If the namespace does not, they are colored the unbound color.
             ;; If use-key? is #f, it adds `to-add' without a key.
             ;; pre: arrow-records is not #f
             (define/private (add-to-range/key text start pre-end to-add key use-key?)
-              (define end (if (= start pre-end) (+ start 1) pre-end))
-              (when (<= 0 start end (send text last-position))
+              ;; Truncate the end of the tooltip region to the final
+              ;; position in the editor. Too-long tooltips occur due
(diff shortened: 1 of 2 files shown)


rfindler:
Thank you.

PR #173, opened 2017-11-25
Added a link to PasteRack

spdegabrielle:
Hi, 
Added a link to PasteRack http://www.pasterack.org/ 
Shot term solution - should be a script in 'script-plugin' (and script-plugin should be in the main distribution IMHO)

Kind regards,

Stephen


spdegabrielle:
No longer required

PR #178, opened 2017-12-04
Allow Syntax-Check Background Color Configuration

shhyou:
This pull request allows users to configure the background color of document variables and unused variables, though the latter is currently disabled in Syntax Check.

A new method `add-text-type` is added to `syncheck-annotations<%>`. This method colors the background of a range of text according to its type. The method`add-background-color` is no longer called by Syntax Check anymore.


drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
```diff
@@ -128,15 +128,36 @@ in order to make the results be platform independent.
...
+   should be drawn on the background of the given range in the editor,
+   when the mouse moves over it.
+
+   This method is usually called by Check Syntax to add background colors to an
+   identifier based on its lexical information. The types @racket['matching-identifiers],
+   @racket['unused-identifier] and @racket['document-identifier] correspond to
+   the color @racket['drracket:syncheck:matching-identifiers],
+   @racket['drracket:syncheck:unused-identifier] and @racket['drracket:syncheck:document-identifier]
+   in the preference, respectively.
+
```
rfindler: Is there a way to add a link to some relevant documentation for the "in the preference" phrase here?

drracket/drracket/info.rkt
```diff
@@ -70,15 +72,17 @@
     (framework:syntax-color:scheme:string ,tol-wob-constants)
     (framework:syntax-color:scheme:text ,tol-wob-constants)
     (framework:syntax-color:scheme:other ,light-blue)
-    (framework:syntax-color:scheme:parenthesis ,dark-plum)
+    (framework:syntax-color:scheme:parenthesis ,sea-green)
     (framework:syntax-color:scheme:symbol ,light-blue)
```
rfindler: Are these color changes intentional? If so, should they perhaps be in a separate commit (with a message explaining what was wrong before)?

rfindler:
I made two comments in the code. Other than that, I usually include docs changes in the same commit as the source change.  Everything else looks good to me.

drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
shhyou: Sounds good. I added a link to <https://docs.racket-lang.org/drracket/color-scheme.html>.

drracket/drracket/info.rkt
shhyou: These changes added color settings for the new entries `'drracket:syncheck:unused-identifier` and `'drracket:syncheck:document-identifier`. I changed some old entries so that matching identifiers and document identifiers (and parenthesis in white-on-black schemes) don't share the same color. Of course, keeping the original color also seems good to me.

shhyou:
I'll rebase the commits after we decide what to do with the color schemes.

rfindler:
Thanks!

PR #220, opened 2018-07-02
Fix test for bitmap button labels.

samth:
See DrDr error here:
http://drdr.racket-lang.org/47742/racket/share/pkgs/drracket-test/tests/drracket/hash-lang-extensions.rkt


rfindler:
Thanks!

PR #229, opened 2018-08-17
Edit inserted message when the evaluation has been terminated

shhyou:
The evaluation could be either killed (by the user) or terminated due to out-of-memory error. For the former case, insert `Evaluation Terminated` instead of `The program ran out of memory.`.

This is an example program:

```
#lang racket/base

(define (g m)
  (sleep 0.5)
  (* 2 (f (+ m 1))))

(define (f n)
  (+ 5 (g (+ n 1))))

(thread
 (let ([main-thread (current-thread)])
   (λ () (sleep 3) (kill-thread main-thread))))

(f 8)
```


rfindler:
Oh, of course! It didn't even occur to me that the stacktrace would show up for killing more generally! Thanks.

PR #232, opened 2018-08-24
Answer a student question about the Run button

ccshan:
https://twitter.com/ShriramKMurthi/status/1032826032590598146


bennn:
This change confused me at first:
1. I keep my interactions window on the right, so just reading the diff I didn't understand the idea
2. Googling for "clean out the bottom" gave very unhelpful results (about dishwashing & other cleaning)

I think it would be more helpful to add a comment about resetting to the main page of these docs; something like the italics below:

> Clicking the Run button evaluates the program in the definitions window, making the program’s definitions available in the interactions window. Given the definition of square as in the figure above, typing (square 2) in the interactions window produces the result 4. _Clicking the Run button again resets the interactions window and re-evaluates the definitions window._

rfindler:
I like Ben's wording. With some `@tech{}`s in the actual code, I presume?

spdegabrielle:
I created an additional pr #388  that adds the text that @bennn suggested.

PR #237, opened 2018-11-02
fix doc of ensure-rep-shown and add check for it

Syntacticlosure:


drracket/drracket/private/unit.rkt
```diff
@@ -2355,6 +2355,9 @@
           (toggle-show/hide-definitions)
           (update-shown)))
       (define/public (ensure-rep-shown rep)
+        (unless (is-a? rep drracket:rep:text<%>)
+           (error 'ensure-rep-shown
```
sorawee, line 2359: The documentation of `error` states:

> When possible, use functions such as `raise-argument-error`, instead, which construct messages that follow Racket’s error message conventions.

So let's use `raise-argument-error` instead.

sorawee: Oh, and thanks to @rfindler for catching this!

rfindler:
Something like this, maybe?

```
(raise-argument-error
  'ensure-rep-shown
  (format "~s" '(is-a?/c drracket:rep:text<%>))
  rep)
```

rfindler:
And also squash the commits. I'm happy to do this if you want, @Syntacticlosure ?

spdegabrielle:
I think this is just a button press now? https://github.blog/2016-04-01-squash-your-commits/

samth:
It is, but it would cause the commit to be authored by @rfindler.

sorawee:
Coauthored I think, just like https://github.com/racket/racket/commit/ee73a550aabd1f6ed72d90f00dddb6915479fc52.

rfindler:
I don't see a commit that fixes the error call. Am I just looking in the wrong place?

I am happy to not have my name on this commit (or to have it; either is fine).

Syntacticlosure:
@rfindler 
It has been a long time since i sent this pull request, so i almost forget it.
It's OK to squash my commits even that it will not be authored by me.

rfindler:
Thank you! And sorry for taking so long to get around to this.

PR #244, opened 2018-12-03
Update 'definition-text-surrogate to accept lists

LeifAndersen:
Before a language's get-info function was expected to return a
module-path? (or #f) for 'definitions-text-surrogate. Now a list is also
acceptable, which can be used for meta-language's that combine their
surrogate with the base language they are extending.

### DIFF
--- a/drracket/drracket/private/in-irl-namespace.rkt
+++ b/drracket/drracket/private/in-irl-namespace.rkt
@@ -55,13 +55,30 @@
          (add-contract 'definitions-text-surrogate
                        (key->contract 'definitions-text-surrogate)
                        (language-get-info 'definitions-text-surrogate #f))))
-  (and surrogate-module
-       (new (add-contract 'definitions-text-surrogate
-                          (implementation?/c
-                           ;; the framework should be shared in the namespace
-                           ;; with this module by the time we get here
-                           (dynamic-require 'framework 'racket:text-mode<%>))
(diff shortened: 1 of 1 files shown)


LeifAndersen:
I've updated the docs. However I can't seem to find any existing tests for 'definitions-text-surrogate.

LeifAndersen:
Ugg...

Thinking about it a bit more, a list _can_ be a module-path? (Like in the case of `'(submod "." hello)`. And with this current design, its impossible to determine if that is supposed to be one module path, or 3. :(

rfindler:
Make a new key that drr will prefer that is always a list.

LeifAndersen:
@rfindler Good idea. I've pushed that.

LeifAndersen:
Any other comments @rfindler ?

rfindler:
Is #f different than the empty list for `definitions-text-surrogate-list`?

The definition of `mode` on line module-language-tools.rkt seems unlikely to be correct as it might be a list and there is a `send` just below.

The docs would be clearer if they said explicitly what order the mixins are mixed in.

This seems unlikely to be correct:

```
If
 @language-info-ref[definitions-text-surrogate] returns
 @racket[#f] then DrRacket tries
 @racket['definitions-text-surrogate]
```

The comment "Will not work with the definitions text surrogate interposition that
#lang allows. Need to deprecate that one" at the top of insulated-read-language.rkt suggests that maybe this is an ill-considered change.

PR #255, opened 2019-01-05
Add "Scribble PDF" button on all platforms.

### LINKED ISSUE #254: "Scribble PDF" button on windows is missing
On MacOS when editing a `#lang scribble/base` DrRacket shows a button "Scribble PDF" to generate a PDF and open it in preview:

scribble-pdf-button

On Windows this button is missing, as already reported here:
https://lists.racket-lang.org/users/archive/2015-March/066280.html

Indeed there is a operating system check to only show the button on MacOS:
https://github.com/racket/drracket/blob/master/drracket/scribble/tools/drracket-buttons.rkt#L70-L79

And opening the generated PDF is done via the `open` shell command which apparently on works on Mac:
https://github.com/racket/drracket/blob/master/drracket/scribble/tools/drracket-buttons.rkt#L63-L64

On the other hand HTML files can be generated and opened by DrRacket across all platforms and are presented to the user via `send-url/file`.

### Proposed Solution

I tried `send-url/file` for PDF files and it works on both Windows and Mac. On Windows it opens the file in the PDF viewer that is configured as default in the system preferences. On Mac it shows the PDF in the default web browser (all of them can render PDFs nowadays).

So I propose the replace the `open` shell command with `send-url/file` and enable the "Scribble PDF" button across all platforms.

I did not tried `send-url/file` on linux but even if it does not work on linux the "Scribble PDF" button should be made available on Windows. Currently I am teaching Racket and many students are running Windows.

laszlokorte:
Fixes #254

Up until now the "Scribble PDF" button was only available on osx.

I tried `send-url/file` on Linux(Fedora, fresh install) and it works opening the PDF in the web browser (firefox). On Windows it shows the PDF in the Default PDF Viewer. The behavior on MacOS is not changed.

I do not know how to build/test DrRacket.

### DIFF
--- a/drracket/scribble/tools/drracket-buttons.rkt
+++ b/drracket/scribble/tools/drracket-buttons.rkt
@@ -28,6 +28,7 @@
    (λ (drs-frame)
      (define fn (send (send drs-frame get-definitions-text) get-filename))
      (define html? (equal? suffix #".html"))
+     (define can-shell-open? (eq? (system-type) 'macosx))
      (cond
        [fn
         (parameterize ([drracket:rep:after-expression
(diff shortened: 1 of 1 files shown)


laszlokorte:
@bennn Thanks you!

As you can see it's working on Windows(7), Mac(Sierra) and Linux(Fedora) now:

bildschirmfoto 2019-01-07 um 20 56 57
bildschirmfoto 2019-01-07 um 20 55 06
bildschirmfoto 2019-01-07 um 20 36 07

PR #265, opened 2019-03-11
fix a Check Syntax internal error

yjqww6:
The following program crashes Check Syntax with an internal error. This PR fixed it.
```
#lang racket
(begin-for-syntax
  (module b racket/base
    (struct A ())))
```


rfindler:
Thank you! I've cherry-picked your commit.

PR #271, opened 2019-04-10
syncheck: Look at 'sub-range-binders properties on disappeared ids

### LINKED ISSUE #267: Check Syntax does not look for 'sub-range-binders properties on identifiers in 'disappeared-binding properties
Here is a macro that expands into `define-syntax-rule` and puts `'sub-range-binders` on the bound identifier:

```racket
#lang racket

(require (for-syntax racket/syntax)
         syntax/parse/define)

(define-syntax-parser define-prefixed
  [(_ x:id)
   #:do [(define x-len (string-length (symbol->string (syntax-e #'x))))]
   #:with prefixed-x (format-id #'x "prefixed-~a" #'x)
   #:with prefixed-x/prop (syntax-property
                           #'prefixed-x
                           'sub-range-binders
                           (vector (syntax-local-introduce #'prefixed-x) 9 x-len 0.5 0.5
                                   (syntax-local-introduce #'x) 0 x-len 0.5 0.5))
   #'(define-syntax-rule (prefixed-x/prop) (void))])
```

This is enough to convince DrRacket to draw a binding arrow when the macro is defined at the top of a module:

Screen Shot 2019-04-09 at 14 15 12

…but if it is defined in a local scope, then it stops working:

Screen Shot 2019-04-09 at 14 15 22

This is because the local syntax binding disappears, which is recorded with `'disappeared-binding`, but Check Syntax doesn’t call `add-binders` or `add-sub-range-binders` on the identifiers inside the `'disappeared-binding` property, just `add-id`.

lexi-lambda:
Fixes #267.

Due to the need to define macros in these tests, the test cases are rather noisy. I don’t know if that is considered a problem or not, but if it is, I can look into what it would take to adjust the test harness so that the macro can be defined in a separate file, so only the macro uses are seen by Check Syntax.

### DIFF
--- a/drracket-test/tests/drracket/syncheck-test.rkt
+++ b/drracket-test/tests/drracket/syncheck-test.rkt
@@ -10,6 +10,7 @@
            racket/list
            racket/file
            racket/set
+           racket/port
            racket/pretty
            racket/gui/base
            framework
@@ -27,7 +28,8 @@
   ;;                        (listof (list number number) (listof string)))
   ;;                        (-> any)
   ;;                        (any -> void?)  -- argument is the result of the setup thunk
-  (define-struct test (line input expected arrows tooltips setup teardown extra-info?) #:transparent)
+  (define-struct test (line input expected arrows tooltips setup teardown extra-files extra-info?)
+    #:transparent)
   (define-struct (dir-test test) () #:transparent)
   
   (define-struct rename-test (line input pos old-name new-name output) #:transparent)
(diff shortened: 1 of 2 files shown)


rfindler:
This looks reasonable to me. I do agree that it would be nice if the macros could be moved out so the test is more focused, but I think that a little explanation of what the test is trying to accomplish in a comment (so people would understand that, when it fails at some point in the future for a dumb reason, they can do the adjustment then)

lexi-lambda:
I have updated this PR with some improvements to the test harness to make it easier to have tests that rely on additional files. Specifically, the test harness now creates a temporary directory instead of just creating a temporary file, which makes it possibly to have reliable file names across test runs without worrying about name conflicts.

rfindler:
Thanks!

PR #284, opened 2019-07-10
Translate a message on the startup screen

alshopov:
The message is "memory limit".  Reuse the limit-memory-limited constant which seems fine
The construction of the message is suboptimal but should be fine for now

Signed-off-by: Alexander Shopov <ash@kambanaria.org>

### DIFF
--- a/drracket/drracket/private/rep.rkt
+++ b/drracket/drracket/private/rep.rkt
@@ -1791,7 +1791,7 @@ TODO
           (insert/delta this (string-append " [" (string-constant custom) "]") dark-green-delta))
         (when custodian-limit
           (insert/delta this 
-                        "; memory limit: "
+                        (format " ~a " (string-constant memory-limit))
                         welcome-delta)
           (insert/delta this
                         (format "~a MB" (floor (/ custodian-limit 1024 1024)))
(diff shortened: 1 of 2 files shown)


rfindler:
Is that string constant defined (for English)?

alshopov:
I think so.

When testing the change with the already released 7.3 the English interface shows:
> Welcome to DrRacket, version 7.3 [3m].
> Language: racket, with debugging [custom]; Limited: 256 MB.

The constant itself is: [`(limit-memory-limited "Limited")`](https://github.com/racket/string-constants/blob/master/string-constants-lib/string-constants/private/english-string-constants.rkt#L1077)

rfindler:
Oh, I see. That constant is designed for a different GUI context (the dialog one gets from the "Racket|Limit Memory..." menu item), so probably it is better to do this via a new string constant (which requires a version number bump in the string-constants library).

alshopov:
Will do tomorrow. I agree the context is somewhat different, I just tried to not introduce a new constant.
On a related note - I will try to extract translatable constants for the macro debugger. That should be OK, right?

rfindler:
Thank you! 

As for the macro stepper, I think probably best to ask @rmculpepper .

alshopov:
This should follow the merge of https://github.com/racket/string-constants/pull/36

rfindler:
Thanks!

PR #323, opened 2019-11-08
scribble: make Scribble PDF button work on 'unix, and simplify for HTML

rrthomas:
Rather than send-url/file on Windows, use start, taking care to supply an
empty first argument (which Windows interprets as a window title).

Use this for HTML too. In fact, xdg-open on GNU/Linux etc., open on macOS,
and start on Windows all work with both files and URLs, so this would work
for http URLs too etc.

Use system* rather than system to avoid quoting problems.

### DIFF
--- a/drracket/scribble/tools/drracket-buttons.rkt
+++ b/drracket/scribble/tools/drracket-buttons.rkt
@@ -56,16 +56,8 @@
                                                                    'scribble/pdf-render)
                                                                'render-mixin)
                                #:xrefs (list xref)))
-                            (define file-to-open (path-replace-suffix fn suffix)) 
-                            (cond
-                              [html?
-                               (send-url/file file-to-open)]
-                              [else
-                               (if (equal? (system-type) 'windows)
-                                   (send-url/file file-to-open)
-                                   (parameterize ([current-input-port (open-input-string "")])
-                                     (system (format "open \"~a\""
-                                                     (path->string file-to-open)))))])))])
+                            (define file-to-open (path-replace-suffix fn suffix))
+                            (send-url/file (path->string file-to-open))))])
           (send drs-frame execute-callback))]
        [else
         (message-box "Scribble" "Cannot render buffer without filename")]))


samth:
I think using `send-url/file` supports opening the user's configured browser, and we should probably keep doing that in general.

rrthomas:
`start`, `open` and `xdg-open` all use the user's configured browser.

samth:
Right, but `send-url/file` also respects browser configuration which you can do inside DrRacket, and we probably don't want to just stop doing that. See https://docs.racket-lang.org/net/sendurl.html#%28def._%28%28lib._net%2Fsendurl..rkt%29._send-url%29%29 for a description.

rfindler:
Thank you!

One fear: what versions of Windows does this work on?

Robby

rrthomas:
@samth all supported versions (it goes back at least to XP).

rrthomas:
@samth: we are opening a local file here, which may be HTML or PDF, not opening a browser, so what we really want is the program the user has configured to open HTML or PDF files. Racket doesn't offer a way to configure that. Really, Racket shouldn't offer a way to configure the browser either, but just use the user's configured browser.

rfindler:
@rrthomas yes! thank you.

samth:
Yes, that sounds like a great idea.

More generally, I agree that we don't really need in-Racket configuration for this, but it exists and should continue to be respected.

rrthomas:
@samth Great, and yes, of course I don't want to upset existing configuration! So I'll come up with a patch that is mostly to netlib/net/sendurl.rkt, where I appear in any case to have previously meddled.

rrthomas:
I've updated the PR, relying on racket/racket#2900 to provide the requisite functionality on Unix (Windows and macOS already work).

rrthomas:
Ping @samth @rfindler : racket/racket#2900 has now landed! Is there anything else I need to do to this PR to make it mergeable?

Issue #235, opened 2018-10-22
Visibility issues with Dr Racket and dark themes


pmplant:
Some elements of the UI, particularly the error message in the lower left, are difficult to see on a dark theme. I could not any way to customise these elements under preferences.

![screenshot at 2018-10-21 16-02-16](https://user-images.githubusercontent.com/20781382/47324727-e2a63200-d615-11e8-8600-a57768f4427c.png)

winny-:
Related racket/racket#2140

aidalgol:
Following on from the discussion in racket/racket#2140, I do not think this is a racket/gui bug, because I get the correct colours with a minimal racket/gui application:
```racket
#lang racket/gui

(let* ((frame (new frame%
                   [label "Dark GTK bug"]))
       (readout (new message%
                     [parent frame]
                     [label "Message"]))
       (button (new button%
                    [parent frame]
                    [label "Button"])))
  (send frame show #t))
```

aidalgol:
I'm having a really hard time finding where in the (DrRacket) code these problem widgets are created, so I'm quite stuck in debugging this myself.

rfindler:
One issue is whether or not `racket/gui` is providing the information that says that dark-on-light mode is enabled.

You are right that the controls that have the wrong font and colors are not standard GUI controls. But they are implemented here: https://github.com/racket/gui/blob/master/gui-lib/mrlib/name-message.rkt

aidalgol:
The problem does seem to be in the MrLib widgets.  The "Name Message" widget in this program will is rendered in black.
```racket
#lang racket/gui

(require mrlib/name-message)

(let* ((frame (new frame%
                   [label "Dark GTK bug"]
                   [width 640]
                   [height 480]))
       (message (new message%
                     [parent frame]
                     [label "Message"]))
       (button (new button%
                    [parent frame]
                    [label "Button"]))
       (name-msg (new name-message%
                      [parent frame]
                      [label "Name Message"])))

  (send frame show #t))
```

aidalgol:
@rfindler Are they all `name-message`s?  Even the line:column readout, and memory usage readout?  What about the "Run", "Debug", etc buttons?

rfindler:
The line:column and memory usage bits are in framework/private/frame, the Run/Debug buttons are mrlib/switchable-button.

sorawee:
The color of error messages of Check Syntax is fixed by f50868afa47921b7e4c074615a37ba918ecf7724.

[...]

PR #332, opened 2019-12-30
Adjusting the "Insert Large Letters..." dialog for Dark mode

rfindler:
Depends on https://github.com/racket/typed-racket/pull/894.


rfindler:
related to #235

Issue #201, opened 2018-04-04
Feature request: automatically delete closing parens of empty s-expressions when automatic parentheses are enabled


lexi-lambda:
I use DrRacket’s “Enable automatic parentheses” feature to automatically insert closing parens when I type opening ones, which I quite like. However, there is one small infelicity about it that I think would be nice if it were changed. Sometimes, I will accidentally type an open paren, then realize I actually did not want it after all. I delete the open paren, but this leaves a dangling close paren—the one automatically inserted by DrRacket—that I must manually delete. This is especially awkward since it is after my cursor, and I must either use forward-delete or the cursor keys to remove it.

Instead, I’d appreciate it if, when automatic parentheses are enabled, DrRacket would delete both parentheses when the open paren is deleted in the text `()`. This would prevent the dangling parens.

Also, for what it’s worth, other editors implement this behavior, such as Atom (via the built-in bracket-matcher plugin, which is enabled by default) and Vim (with the commonly-used autoclose plugin, which is the equivalent of DrRacket’s automatic parentheses mode).

rfindler:
Is it better to always delete the close when the open is deleted, or only
when it is an empty sexp?

Robby

lexi-lambda:
I think it should only delete the close paren when it is an empty s-exp. That’s what the other editors I listed do, so they probably do it that way for a reason, and it’s probably better for consistency. If it deleted all matching parens whenever a paren was deleted, that would be interesting, but I think that’s pretty dramatically different from what automatic parentheses currently does.

rfindler:
Okay. The other question is: does it do this only when one types the delete key, or when the paren is deleted by any means? (cmd/ctl-x or who knows what other myriad ways there are to delete things)

lexi-lambda:
Anecdotally, in both Atom and vim, it *only* happens when the user deletes the open paren using the backspace key. If they delete it any other way, including forward-delete or vim normal-mode keybindings (including if using vim-mode in Atom), it doesn’t delete the close paren.

[...]

PR #343, opened 2020-02-09
updated docs for auto-delete feature

bzyeung:
Documentation related to #201 and racket/gui#162

This is a fairly minimal update.  Should I also update the language in the Editing with Parentheses section?


rfindler:
Thanks!

PR #346, opened 2020-02-16
Extract new translatable constants, bump dependency

alshopov:
Signed-off-by: Alexander Shopov <ash@kambanaria.org>


alshopov:
Land after https://github.com/racket/string-constants/pull/43

PR #357, opened 2020-03-19
Drop `#f` frames in stack checkpoint test.

samth:
The `#f` frames are already dropped in `cut-stack-at-checkpoint`, causing
   this test to fail on Racket CS which has some extra `#f` frames at the
   beginning.

   Also, make tests more informative.


rfindler:
Thanks!

PR #359, opened 2020-03-22
Repair `drracket:debug:error-display-handler/stacktrace`.

samth:
Commit 4b3cff1abfa69 changed (perhaps unintentionally) the contract
for `error-display-handler/stacktrace`. The previous documented
contract expected `(or/c #f list?)` but the new code expects a
`viewable-stack?`.

This changes the two clients in this repository, but there might
be others elsewhere.

`tests/drracket/synchek-test`, which was broken by the change,
passes after this commit.

### DIFF
--- a/drracket/drracket/private/debug.rkt
+++ b/drracket/drracket/private/debug.rkt
@@ -336,7 +336,7 @@
               (orig-error-display-handler msg exn)])))))
     debug-error-display-handler)
   
-  ;; error-display-handler/stacktrace : string any (listof srcloc) -> void
+  ;; error-display-handler/stacktrace : string any (or/c #f viewable-stack? (listof srcloc)) -> void
   ;; =User=
   (define (error-display-handler/stacktrace 
            msg exn 
@@ -347,7 +347,8 @@
                                            (send rep get-definitions-text)))])
     (define stack1
       (cond
-        [pre-stack pre-stack]
+        [(viewable-stack? pre-stack) pre-stack]
+        [(list? pre-stack) (srclocs->viewable-stack pre-stack (list ints defs))]
         [(and (exn? exn)
               (continuation-mark-set? (exn-continuation-marks exn)))
          (cms->errortrace-viewable-stack (exn-continuation-marks exn)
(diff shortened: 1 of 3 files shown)


rfindler:
The intention was that viewable stacks are accepted as arguments (if available and they would be available only internally) but that other things are also accepted and converted. I'm not 100% sure this is possible but that was the intention (sorry I didn't look into this failing test case for so long).

samth:
Ok, I'll see if I can write the conversion code.

samth:
Ok, I think it seems right now. Note that I also fixed two places where `empty-viewable-stacktrace` was used instead of `(empty-viewable-stacktrace)`.

rfindler:
This is the kind of change that makes sense to me.

Except that I don't think that the tool-lib contract should change. That contract isn't enforced internally only on tools and probably tools should actually obey it. I didn't try to run any code, tho, so is there something wrong with that?

rfindler:
Also, the interesting editors argument of the call to `srclocs->viewable-stack` should be a list containing the defs and ints arguments. 

And yes, `empty-viewable-stack` used to be a constant in an earlier version of this code (probably that wasn't pushed to master) and then viewable stacks got some state, so that changed into a thunk. That's definitely a good change.

samth:
Ok, that's done.

rfindler:
Thanks! Shall I merge?

samth:
Sure.

samth:
I went ahead and merged.

rfindler:
Thanks!

samth:
Looks like the last change wasn't tested enough: using `(list ints defs)` causes the following error:

```
exception raised by error display handler: send: target is not an object
  target: #f
  method name: get-edition-number; original exception raised: read-syntax: expected a `)` to close `(`
```

@rfindler Any thoughts on what's happening?

samth:
Further investigation:

The problem is that `(drracket:rep:current-rep)` is producing `#f`, causing the default argument value to be `#f` for `ints` and `defs`. However, the call site in `syncheck/gui.rkt` supplies the `#:definitions-text` keyword. If I change it to supply both, the tests pass again.

However, it seems like this could be a problem for other tools that call this function, so I think some sort of validation is needed.

PR #360, opened 2020-03-24
make-traversal takes a directory not a full file path name

greghendershott:


rfindler:
Thanks!

Can you say how you found this problem? (I would like to make a test case.)

rfindler:
Ah, I see one way. Nevermind.

PR #401, opened 2020-07-31
clarify docs for add-online-expansion-handler

maueroats:
rephrase docstring for `add-online-expansion-handler` and `add-online-expansion-monitor`

### DIFF
--- a/drracket/drracket/tool-lib.rkt
+++ b/drracket/drracket/tool-lib.rkt
@@ -238,8 +238,8 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     (See also @racket[drracket:module-language-tools:add-online-expansion-monitor].)
     
-    The first two arguments name a procedure in a module that is loaded by 
-    @racket[dynamic-require] is specially designed separate @racket[place].
+    The procedure @racket[id] from @racket[mod-path] is loaded by 
+    @racket[dynamic-require] into a specially designed separate @racket[place].
     When DrRacket detects that
     the editor has been modified, it sends the contents of the editor over to
     that separate place, @racket[expand]s the program there, and then supplies
@@ -297,7 +297,7 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     
     Like @racket[drracket:module-language-tools:add-online-expansion-handler], 
-    the first two arguments name a procedure that is called in the separate place
+    the first two arguments specify a procedure that is called in the separate place
     designated for expansion.
     
     The procedure is called before expansion starts and once it returns, expansion
--- a/drracket/drracket/tool-lib.rkt
+++ b/drracket/drracket/tool-lib.rkt
@@ -238,7 +238,7 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     (See also @racket[drracket:module-language-tools:add-online-expansion-monitor].)
     
-    The procedure @racket[_id] from @racket[mod-path] is loaded by
+    The procedure @racket[id] from @racket[mod-path] is loaded by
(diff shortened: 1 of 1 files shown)


drracket/drracket/tool-lib.rkt
```diff
@@ -238,8 +238,8 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     (See also @racket[drracket:module-language-tools:add-online-expansion-monitor].)
     
-    The first two arguments name a procedure in a module that is loaded by 
-    @racket[dynamic-require] is specially designed separate @racket[place].
+    The procedure @racket[id] from @racket[mod-path] is loaded by
```
rfindler, line 241: This needs to be @racket[_id] to get the right typesetting.

drracket/drracket/tool-lib.rkt
```diff
@@ -238,8 +238,8 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     (See also @racket[drracket:module-language-tools:add-online-expansion-monitor].)
     
-    The first two arguments name a procedure in a module that is loaded by 
-    @racket[dynamic-require] is specially designed separate @racket[place].
+    The procedure @racket[id] from @racket[mod-path] is loaded by 
+    @racket[dynamic-require] into a specially designed separate @racket[place].
```
rfindler, line 242: Maybe something like `@tech[#:doc '(lib "scribblings/reference/reference.scrbl")]{place}` would be better than `@racket[place]`?

rfindler:
I made two little comments -- happy to make those changes myself if you'd prefer and then just merge.

Thanks!

samth:
@rfindler if you wan to make an actual edit to the code, you can click the `+-` button while you're typing the comment, and offer a suggestion which can be incorporated directly in the web interface.

drracket/drracket/tool-lib.rkt
samth:
```suggestion
    The procedure @racket[_id] from @racket[mod-path] is loaded by
```

drracket/drracket/tool-lib.rkt
rfindler:
```suggestion
    @racket[dynamic-require] into a specially designed separate @tech[#:doc '(lib "scribblings/reference/reference.scrbl")]{place}.
```

rfindler:
Nice! Thanks, @samth . (Someone should build the docs and look at the output before pushing probably, tho.)

drracket/drracket/tool-lib.rkt
samth:
```suggestion
    @racket[dynamic-require] into a specially designed separate @tech[#:doc '(lib "scribblings/reference/reference.scrbl")]{place}.
```

maueroats:
1. Confirmed that the docs render ok.
2. I see what `@racket[_id]` is supposed to do, but in my local install, both `@racket[_id]` and `@racket[id]` produce the same HTML (a `span` with `class="RktVar"`). If you think all of the functions' formal parameters should have underscores in front of their symbols, then this PR could get bigger to make the edits. I'm not undoing your edit, just wondering.

rfindler:
Oh, my mistake on the `_`! I was confused -- `id` is already part of the reader so the rendering is automatic. Sorry about that.

drracket/drracket/tool-lib.rkt
```diff
@@ -238,8 +238,8 @@ all of the names in the tools library, for use defining keybindings
   @{Registers a pair of procedures with DrRacket's online expansion machinery. 
     (See also @racket[drracket:module-language-tools:add-online-expansion-monitor].)
     
-    The first two arguments name a procedure in a module that is loaded by 
-    @racket[dynamic-require] is specially designed separate @racket[place].
+    The procedure @racket[_id] from @racket[mod-path] is loaded by
```
rfindler, line 241:
```suggestion
    The procedure @racket[id] from @racket[mod-path] is loaded by
```

PR #412, opened 2020-09-03
Fix another typo

sorawee:


sorawee:
@bogdanp: do you know why 

```
  with:
    architecture: x64
    distribution: full
    variant: CS
    version: current
```

results in `##[error]Unexpected HTTP response: 404`? Would bumping the version up to 0.10 solves the problem?

Bogdanp:
@sorawee yep, bumping the version will fix the problem.

Details [here](https://github.com/Bogdanp/setup-racket/pull/12) and [here](https://groups.google.com/g/racket-users/c/k_o3Ug-lK_s/m/HwhCtbC6BQAJ).

PR #418, opened 2020-09-10
add relative path support to make-traversals

sorawee:
The current requirement of path in make-traversal is that it is
either #f or a path-string? However, because we parameterize
current-load-relative-directory to be this path, providing
any non complete-path? will result in a contract error.
This PR thus strengthens the requirement so that the provided path
must be complete-path?

The PR also fixes the show-content procedure so that
it satisfies the contract. Note that we need additional
parameterization of current-load-relative-directory so that
macro expansion can work correctly on relative path source like:

```
;; main.rkt
(require drracket/check-syntax)
(show-content "test/a.rkt")

;; test/a.rkt
(require "b.rkt")

;; test/b.rkt
1
```

Also cleanup description about `add-id-set`, since the method is no longer called.

### DIFF
--- a/drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
+++ b/drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
@@ -41,8 +41,7 @@ that are exposed via Racket APIs to be used with other editors.
  the vector are the arguments passed to the method.
  (Note that this procedure does not account for the callback
  procedures present in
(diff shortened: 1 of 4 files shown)


rfindler:
Probably better if we do `(build-path (current-directory) ...)` when `complete-path?` returns #f for the input (and document that) instead of strengthening the contract.

sorawee:
Ping, any thoughts on this? It's been a month.

PR #426, opened 2020-10-11
Enable markup support for the repl editor:

mikesperber:
- add markup to the modules shared between client and DrRacket.
- override find-editor from frame:text<%> for DrRacket.

NOTE: Needs to be applied together with https://github.com/racket/gui/pull/201


drracket/drracket/private/unit.rkt
```diff
@@ -4711,7 +4711,17 @@
...
+      (define/override (find-editor predicate)
+        (cond
+          [(findf predicate (map (lambda (tab) (send tab get-defs))
+                                 tabs))
+           => (lambda (editor)
+                editor)]
+          [(predicate interactions-text)
+           interactions-text]
+          [else #f]))
+
```
rfindler, line 4724: This look looks like it is searching in every definitions text but only one interactions text. Is that intended?

mikesperber:
Good point - that's a holdover from the old code. Fixed.

rfindler:
Maybe something else needs to be done to update the pull request? I'm not seeing a change here.

mikesperber:
@rfindler Sorry, one of these days I'll figure out git. Better now?

PR #449, opened 2021-01-06
Handle drscheme:opt-out-toolbar-buttons correctly

mikesperber:
Either of drscheme:opt-out-toolbar-buttons /
drracket:opt-out-toolbar-buttons can be used to turn off the opt-out
buttons.  The way it was implemented, you needed both.  Unbreak.

### DIFF
--- a/drracket/drracket/private/module-language-tools.rkt
+++ b/drracket/drracket/private/module-language-tools.rkt
@@ -356,8 +356,10 @@
          (or (call-read-language the-irl 'drracket:toolbar-buttons #f)
              (call-read-language the-irl 'drscheme:toolbar-buttons #f))
          
-         (or (call-read-language the-irl 'drracket:opt-out-toolbar-buttons '())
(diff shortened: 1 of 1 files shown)


rfindler:
I think that it needs to be something more like "replace `#f` with `'()` and then do an append". That is, if drr is `#f`  and drs is a list of useful things, we want the list of useful things.

rfindler:
Oh actually, I think we want to choose *either* drr or drs (preferring the drr ones).

mikesperber:
> Oh actually, I think we want to choose _either_ drr or drs (preferring the drr ones).

The "old" code in, say, `htdp/bsl/reader.rkt`, knows only the `drscheme:` prefix and expects `drscheme:opt-out-toolbar-buttons` -> `#f` to remove the extra buttons. If this isn't expected to work, `drscheme:opt-out-toolbar-buttons` should be removed entirely.

rfindler:
Oh, I see. Righto!

PR #453, opened 2021-01-20
fix gui-debugger annotator

rmculpepper:
Use quote-syntax instead of syntax to save variable name.
Fixes racket/racket#3633.


rfindler:
Let me find a little time to make a test case here (at the level of the debugger). I'm happy to make one, but probably not today.

rfindler:
I've pushed this commit to master (and added a test case in another commit).

PR #460, opened 2021-02-03
Fix error highlighting when background expansion is enabled

### LINKED ISSUE #455: Syntax Source is Different when Running Inside DrRacket [cs]
I am not sure if this is a DrRacket issue or a Racket issue.

Version: (Dr)Racket 7.9.0.22 [cs]

Example program:
```
#lang racket/base
(syntax-source (quote-syntax here))
```

When running from the command line, the result is an absolute path:
```
$ racket stxsrc.rkt 
#<path:/Volumes/ramdisk/stxsrc.rkt>
```

When running inside DrRacket [cs] **without errortrace** but **with background expansion**, the result is a relative path:
```
Welcome to DrRacket, version 7.9.0.22 [cs].
Language: racket/base [custom]; memory limit: 512 MB.
#<path:stxsrc.rkt>
> 
```

This difference does not show up in Racket v7.9 [bc]. Subsequently, the path difference affects error location highlighting because `set-error-ranges` relies on `port-name-matches?`.
(related to https://github.com/racket/racket/issues/3325#issuecomment-752354258)
https://github.com/racket/drracket/blob/6cda6a7b1738b0be624f96aba10c48b1e5adbe4d/drracket/drracket/private/rep.rkt#L637-L644.

sorawee:
Currently, under the following conditions:

1. Errortrace is disabled
2. Background expansion is enabled, and it has finished expanding the program
3. The editor is unsaved (not named).

DrRacket will fail to highlight error. This is because the expanding
place uses a dummy srcloc "unsaved editor" which does not match the port
name of any editors, so the editors refuse to recognize the srcloc.

Because the port name of an editor is an uninterned symbol, passing it
to the expanding place doesn't look like a good idea. Instead, we
continue to use the dummy srcloc and adjust its source to be the port
name later.

Fixes #455

### DIFF
--- a/drracket/drracket/private/debug.rkt
+++ b/drracket/drracket/private/debug.rkt
@@ -358,7 +358,18 @@
                                        #:share-cache stack1)
           (empty-viewable-stack)))
     (define port-name-matches-cache (make-hasheq))
-    (define src-locs (get-exn-source-locs defs exn stack1 stack2))
+    (define raw-src-locs (get-exn-source-locs defs exn stack1 stack2))
+    (define src-locs
+      (for/list ([src-loc (in-list raw-src-locs)])
+        (match-define (srcloc src line col pos span) src-loc)
+        (srcloc (if (and defs (equal? "unsaved editor" src))
+                    ;; if source is "unsaved editor", it comes from an
+                    ;; expanding place of an unsaved editor. We need to fix
+                    ;; its source to be the port name for error highlighting
+                    ;; to work correctly.
+                    (send defs get-port-name)
+                    src)
+                line col pos span)))
 
     (print-planet-icon-to-stderr exn)
     (unless (exn:fail:user? exn)


sorawee:
Looking at this again, I'm not convinced at all that this fix is at the right place. Maybe the fix should be in `get-exn-source-locs`? Or maybe the fix should be correcting all srcloc under `exn`, `stack1`, and `stack2`? I have no idea what's the right thing to do.

rfindler:
I'm getting a bit lost on the code that's actually in the (unsaved) definitions window. Is it the [example in these docs](https://docs.racket-lang.org/reference/exns.html#%28def._%28%28quote._~23~25kernel%29._prop~3aexn~3asrclocs%29%29) is the one that we're talking about?

sorawee:
Yes.

rfindler:
The problematic case seems to be when DrRacket sends the already-compiled program back to be evaluated, i.e. when [compiled-bytes](https://github.com/racket/drracket/blob/master/drracket/drracket/private/expanding-place.rkt#L278) isn't `#f`.

rfindler:
Given that this funny value that's meant to represent the source location of an unsaved definitions window can just be embedded into the compiled object, it seems like this has to be made more "official", perhaps via a specific prefab struct that gets documented somewhere and various places adjusted to work with that instead of the "hidden" unique value.

sorawee:
Ideally, running with or without background expansion (that is, through expanding place or not) should produce the same result.

When an unsaved editor is run without background expansion, it uses an [uninterned symbol `unsaved-editor`](https://github.com/racket/gui/blob/master/gui-lib/framework/private/text-basic.rkt#L540) as the source location. Would the protocol that you propose affect this too?

IIUC, the reason it is uninterned is to guarantee uniqueness of port name of each editor. But prefab structure would kill this guarantee, right?

rfindler:
> Ideally, running with or without background expansion (that is, through expanding place or not) should produce the same result.

Yes.

> When an unsaved editor is run without background expansion, it uses an [uninterned symbol `unsaved-editor`](https://github.com/racket/gui/blob/master/gui-lib/framework/private/text-basic.rkt#L540) as the source location. Would the protocol that you propose affect this too?

Yes.

> IIUC, the reason it is uninterned is to guarantee uniqueness of port name of each editor. But prefab structure would kill this guarantee, right?

Yes.

There doesn't seem to be a way to "protect" the identity of the definitions window in the sources so we'd need to reveal more information (which is what I meant by my "coded" comment about documentation).

sorawee:
I will close this PR then, since its approach is clearly wrong. I might revisit this bug later if it's not already been fixed.

rfindler:
The more I'm thinking about this, the more I think that it is fine to be able to "forge" being in the definitions window as the source. After all, you can already "forge" arbitrary paths. So I'm thinking that declaraing that some specific prefab (with some integer or pair of integes in it) will just be interpreted as belonging to a specific definitions window is probably the way to go. There is some question as to what that means when you save one of these and restart drracket and manage to get an old one, but perhaps that's just life.

rfindler:
Hm. I see that `get-port-name` is already part of a public API. So maybe what I have to do is just encode more structure into that symbol and stop it from being a unique symbol.

PR #462, opened 2021-02-04
backtrace: fix empty backtrace on subsequent activation

### LINKED ISSUE #461: Loading backtrace for the second time produces empty backtrace
Steps to reproduce:

1. Run `#lang racket (/ 1 0)`.
2. Click at the backtrace button. You should see something like:

Screen Shot 2021-02-03 at 3 10 57 PM

3. Close the window, and click at the backtrace button again. Now, the backtrace is empty!

Screen Shot 2021-02-03 at 3 11 06 PM

sorawee:
Backtrace display mutates viewable stacks. Therefore, we should copy
them first so that subsequent activations still produce correct result.

Fixes #461

### DIFF
--- a/drracket/drracket/private/debug.rkt
+++ b/drracket/drracket/private/debug.rkt
@@ -542,10 +542,9 @@
        (send note set-stacks viewable-stack1 viewable-stack2)
        (send note set-callback
              (λ (snp)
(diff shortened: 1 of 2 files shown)


rfindler:
Thanks!

PR #464, opened 2021-02-04
backtrace: fix font size inconsistency

### LINKED ISSUE #463: font size inconsistency
Consider:

```
#lang racket
(add1 (let loop ([n 5])
        (cond
          [(zero? n) (/ 1 0)]
          [else (add1 (loop (sub1 n)))])))
```

The backtrace displays:

Screen Shot 2021-02-03 at 8 49 41 PM

Notice that 
- the size of the first srcloc is different from the size of other srclocs
- the size of most srclocs are different from the size of "skip duplicate frames"
- srclocs have a space between source name and line&column, which is not how srclocs are usually formatted.

It's unclear what is the desired outcome, but it definitely should not be inconsistent like this.

sorawee:
Also adjust srcloc display to be consistent with
how it's usually displayed.

Fixes #463

### DIFF
--- a/drracket/drracket/private/debug.rkt
+++ b/drracket/drracket/private/debug.rkt
@@ -872,8 +872,14 @@
     (send text lock #t)
     (send text hide-caret #t)
     (send current-backtrace-window show #t))
-  
-  
+
+  (define (change-regular-style text start-pos end-pos)
+    (send text change-style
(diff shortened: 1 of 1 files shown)


sorawee:
Screenshot:

Screen Shot 2021-02-04 at 7 29 49 AM

rfindler:
Thanks!

PR #480, opened 2021-04-07
Respect the new print-value-columns parameter.

mikesperber:


rfindler:
I was thinking that we'd replace the use of `drracket:module-language:drracket-determined-width` with `print-value-columns`. So DrRacket would set the value of it where it's currently setting `drracket-determined-width` and then look up only `print-value-columns in `global-port-print-handler`.

Also, I wonder if Racket Mode could do something similar? @greghendershott

greghendershott:
I lack some context.

- What is `drracket:module-language:drracket-determined-width`? A "static" parameter whose value is set by... what? Or a dynamic function that reflects the window width? I thought you [mentioned something like the latter](https://github.com/racket/racket/pull/3754#issuecomment-812879942) but I can't seem to find code doing that, at least from grepping drracket source.

- From https://github.com/racket/racket/pull/3754/files I get the sense that this is supposed to be an alternative to setting `pretty-print-display-width`. The docs say the "consumer" of this is supposed to be pretty-printing REPLs. But the docs don't say who is supposed to "produce" this. Is the idea that say a lang implementation would set this, but a "normal" end user program probably should not?  (If so, it would help if its docs state that. I feel like, overall, when it comes to printing, there are a lot of knobs to twist, and there's decent reference documentation about each one, but not a lot of guide info about intended use.)

Meanwhile:

- I _think_ the TL;DR is you're saying that Racket Mode should check this new `print-value-columns` value (if it can `dynamic-require` that in a new-enough Racket), and if an integer, use it. Else default to the dynamic window-width thing it does.

- And the other TL;DR is you recommend Racket Mode use a global print handler for the REPL output port, and not touch things like `pretty-print` hooks and params for other ports. Which is a to-do for me to do someday.

?

greghendershott:
Oh also: When grepping, I noticed in `module-language.rkt` this:

```racket
      (define/private (do-print value settings port width)
        (parameterize ([drracket-determined-width width])
          (print value port))
        (newline port))
```

Should this also consider `print-value-columns`? If not, why not?

rfindler:
Ah, sorry! DrRacket (when printing out values in the interactions window) takes the width of the window into account. So if you evaluation `(make-list 40 #f)` with a wide interactions window, it'll show up on one line and if you do it with a narrow window, it'll appear on multiple lines. The way this happens is that DrRacket computes the width of the window and then stashes it, only to bring it back later in DrRacket's default value of the `global-port-print-handler`, and use it to control the width. (DrRacket internally is using the pretty printer to print values instead of the regular one, but aside from seeing extra newlines in the output, is trying to hide this fact from the program itself [*].)

As for the specifics of which code should look at it, I am not completely sure; I think some experimentation and test suites will be required to sort everything out properly. 

[*]: the goal is that, if the program uses pretty print for its own purposes and goes around setting various parameters coming from the pretty-print library, those settings will affect the printer exactly as much as it does in command-line racket (i.e., not at all).

rfindler:
> Oh also: When grepping, I noticed in `module-language.rkt` this:
> 
> ```racket
>       (define/private (do-print value settings port width)
>         (parameterize ([drracket-determined-width width])
>           (print value port))
>         (newline port))
> ```
> 
> Should this also consider `print-value-columns`? If not, why not?

I think the general principle here is that we should have all language-based configuration actually be happening in the source code, not in the IDE. This particular method is part of the apparatus that controls printing via settings in the IDE (and was written before I really grokked the first sentence in this paragraph). So the answer is .... maybe? Backwards compatibility becomes the big concern here and it can be tricky to suss out what's going to break what when we change things like that code. I rely on the DrRacket test suites to help me out.....

rfindler:
And I see that I forgot about the constraint that you have to deal with many different versions of Racket. That does seem to be tricky.

greghendershott:
Re `module-language.rkt`: That makes sense; I didn't appreciate the role of that file.

> And I see that I forgot about the constraint that you have to deal with many different versions of Racket. That does seem to be tricky.

That aspect should be straightforward for this `with-handlers` + `dynamic-require` defaulting to `(make-parameter #f)`.

(Things like `racket/class` seems trickier; like someday if `syncheck` gets a new method. Haven't had to handle that, so far. Will cross that bridge if/when we come to it.)

Anyway I'll take a look at isolating the Racket Mode output similar to Dr Racket, and using this new parameter when present.

mikesperber:
> I was thinking that we'd replace the use of `drracket:module-language:drracket-determined-width` with `print-value-columns`. So DrRacket would set the value of it where it's currently setting `drracket-determined-width` and then look up only `print-value-columns in `global-port-print-handler`.

I agree with that, but I suggest not doing that right before the release.

PR #499, opened 2021-07-23
Avoid phase+space error.

samth:
Fixes http://drdr.racket-lang.org/57872/cs/racket/share/pkgs/drracket-test/tests/drracket/syncheck-test.rkt

cc @rfindler

### DIFF
--- a/drracket-tool-lib/drracket/private/syncheck/traversals.rkt
+++ b/drracket-tool-lib/drracket/private/syncheck/traversals.rkt
@@ -979,6 +979,8 @@
   (define binding (identifier-binding var phase-level))
   (and (pair? binding)
        (or (not (number? phase-level))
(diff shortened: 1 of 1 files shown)


rfindler:
This doesn't really fix anything, per se. I'll merge it with a different commit message; hope that's okay.

samth:
That's fine with me.

rfindler:
Thanks, @samth ! Sorry to be slow here.

It also appears that won't completely quiesce drdr. Looks like something improved somewhere! (So a different test case in that file also fails.)

rfindler:
Oops, I got it backwards. It actually appears to be a regression for this program:

```
#lang racket/base
(require (for-syntax racket/base))
(begin-for-syntax (local-require racket/list) first)
```

The arrow from racket/list to first went away at some point.

PR #500, opened 2021-08-02
Have note% implement srcloc-special<%>.

mikesperber:
This way, users of simple-tree-text-markup/port can generate links to
srclocs represented by the note% snips.

This goes back to a discussion @rfindler and I have had. Should be viewed as a first draft.


rfindler:
LGTM

PR #511, opened 2021-09-16
Add an option to directly send the URL without trampolines

shhyou:
This PR adds an option in the browser preference to send URLs directly without going through trampolines. Comments welcome!

- [x] Draw the label of the checkbox from `string-constants`
- [ ] docs
- [x] Code cleanup

### DIFF
--- a/drracket/browser/external.rkt
+++ b/drracket/browser/external.rkt
@@ -6,6 +6,7 @@
            racket/list
            racket/match
            (prefix-in raw: net/sendurl)
+           net/sendurl/preferences
            net/url
            (prefix-in fw: framework))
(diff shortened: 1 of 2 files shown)


rfindler:
This looks okay to me, except the string constants issue and the docs. There is a document that has all of the checkboxes and menu items in DrRacket with some explanation of them [here](https://github.com/racket/drracket/blob/master/drracket/scribblings/drracket/interface-ref.scrbl). Maybe adding a note there with a index entry or something like "documentation on chromebooks" or similar might work?

shhyou:
Thanks! I forgot about the docs.

Issue #517, opened 2021-10-06
Split off module browser from `drracket-tool-lib` to a new `drracket-tool-lib-gui` package?


greghendershott:
I've been looking at packages that depend on `gui-lib` when maybe they need/ought not to.

One is `drracket-tool-lib`. However "most" of it doesn't use `racket/gui/base`. I think the only portions that do are two modules:

- https://github.com/racket/drracket/blob/master/drracket-tool-lib/drracket/module-browser.rkt
- https://github.com/racket/drracket/blob/master/drracket-tool-lib/drracket/private/standalone-module-browser.rkt

Would it be possible move those two files into their own package, called (say) `drracket-tool-lib-gui`, and adjust the `drracket-tool` meta package?

IIUC the files would move within this git repo, and get their own package `info.rkt` and node on `pkgs.racket-lang.org` --- but the collection paths would remain the same so they'd still end up installed in the same place and nothing would break. I think. Does that seem correct?

rfindler:
Thanks, @greghendershott

PR #520, opened 2021-10-20
Move non-GUI parts of drracket-tool-lib to new drracket-tool-text-lib

greghendershott:
This is to follow up on issue #517.

Not sure if the ball was in my court. I wanted to (hopefully) help with at least a draft PR to be more specific about the proposal. Although this isn't urgent, for me, I didn't want it to fall between the cracks.

---

This PR moves the non-GUI parts of `drracket-tool-lib` to a new `drracket-tool-text-lib` directory, and in the former's `info.rkt` adds a dep on the latter.

The non-GUI parts turn out to be effectively "everything except the GUI module browser".

**WARNING**: I'm not sure I understand the correct roll-out steps here.

This PR assumes the steps are simply:

1. Create a new `drracket-tool-text-lib` package on pkgs.racket-lang.org (which should be under @rfindler 's account there, not mine, correct?).
2. Merge this PR.

But would that leave an inconsistent window?

Instead should it be something like the following?

1. Do one PR which simply _copies_ the non-GUI code to the new directory for the new package (temporarily leaving copies of the non-GUI code in _both_).
2. Create the new package catalog entry.
3. Do another PR which _deletes_ the non-GUI code from `drracket-tool-lib` and updates its `info.rkt` to depend on `drracket-tool-text-lib` (of course leaving the copy there).


rfindler:
I've merged the commit. Thanks!

greghendershott:
It looks like `drracket-tool-text-lib` has a [build failure](https://pkg-build.racket-lang.org/server/built/fail/drracket-tool-text-lib.txt):

```
The time is now Saturday, October 30th, 2021 6:38:16pm
(/usr/bin/env DISPLAY=:1 PLT_PKG_BUILD_SERVICE=1 PLTUSERHOME=/home/root//user PLT_PKG_BUILD_SERVICE=1 CI=true PLTSTDERR=debug@pkg error PLT_INFO_ALLOW_VARS=;PLT_PKG_BUILD_SERVICE PLTCOMPILEDROOTS=/home/root//zo: /usr/bin/xvfb-run -n 1 /bin/sh -c cd "/home/root/"/racket && bin/racket -MCR "/home/root/"/zo: -l- raco pkg install -u --auto drracket-tool-text-lib)
Resolved "drracket-tool-text-lib" via file:///home/root//catalogs/archive/catalog
pkg: catalog response: #hash((author . "robby@racket-lang.org") (checksum . "33fb77f5fbfc42db0485d1b1786a2cee16c2ea96") (dependencies . (("base" #:version "6.2.900.10") ("scribble-lib") ("string-constants-lib" #:version "1.12") ("scribble-lib") ("racket-index") ("at-exp-lib") ("rackunit-lib"))) (description . "Code implementing non-GUI programmatic interfaces to some IDE tools that DrRacket supports") (modules . ((lib "drracket/check-syntax.rkt") (lib "drracket/private/syncheck/annotate.rkt") (lib "drracket/private/syncheck/syncheck-intf.rkt") (lib "drracket/private/syncheck/syncheck-local-member-names.rkt") (lib "drracket/private/syncheck/xref.rkt") (lib "drracket/private/syncheck/traversals.rkt") (lib "drracket/private/syncheck/colors.rkt") (lib "drracket/find-module-path-completions.rkt") (lib "drracket/private/syncheck/contract-traversal.rkt"))) (name . "drracket-tool-text-lib") (ring . 2) (source . "file:///home/root/catalogs/archive/pkgs/drracket-tool-text-lib.zip") (tags . ()))
racket pkg install: packages conflict
  package: drracket-tool-text-lib
  package: drracket-tool-lib
  module path: drracket/private/syncheck/contract-traversal
The time is now Saturday, October 30th, 2021 6:38:19pm
```

If I understand the message, it's because it thinks `drracket-tool-lib` already installed `drracket/private/syncheck/contract-traversal`?

Is that because `drracket-tool-lib` did not get rebuilt and was used from a pre-existing cache?  As I type this, its [package page](https://pkgs.racket-lang.org/package/drracket-tool-lib) has a blank (no link) for "Most recent build results", which seems to support this theory (?).

greghendershott:
Speaking of package pages, [drracket-tool-lib](https://pkgs.racket-lang.org/package/drracket-tool-lib) has `Ring = 0` and `Tags = main-distribution`.

Whereas [drracket-tool-text-lib](https://pkgs.racket-lang.org/package/drracket-tool-text-lib) has `Ring = 2` and blank `Tags`.

Since the former depends on the latter, this seems wrong?  Should `drracket-tool-text-lib` also be `Ring = 0` and `Tags = main-distribution`?

Although I don't know exactly how all this works, it seems this could be a reason for the build problem in the previous comment?

jeapostrophe:
I believe that on @mflatt has the answer to why the builds are failing. One hypothesis is that the pkg-build is not using master, but is using the previous release. As far as the ring number, etc, while I do manage the ring status of packages, the ones that are part of `main-distribution` are controlled by another entry point that is part of the release process. This is further evidence for me that we need to wait until after the release for this to happen.

rfindler:
I'm not completely following, @jeapostrophe -- this change should be post release in the sense that the release branches are all already created and this change is not part of them. Are you saying that any change that affects which pkgs depend on which ones needs to be synchronized with the release somehow?

jeapostrophe:
I believe that this question is "A. There's a build failure; B. The package info is weird; Why?"

For B, the answer is that `main-distribution` is controlled via a special path called `api/upload` which happens as part of releases --- https://github.com/racket/pkg-index/blob/master/official/dynamic.rkt#L78 --- by the `libs` repo --- https://github.com/racket/libs/blob/master/pack-and-upload.rkt or maybe `pkg-push` --- https://github.com/racket/pkg-push/blob/master/push-catalog.rkt or maybe `release-catalog` --- https://github.com/racket/release-catalog

For A, I think the answer is that the "release catalog" is used to initialize `pkg-build` --- https://github.com/racket/pkg-build --- rather than `master`, so it won't see things that haven't been in a release yet.

rfindler:
A: Oh, I think I see. So the fact that there is a failure at the link @greghendershott mentioned is more of a problem that it noticed that the pkg existed (because I had to register it and cannot register it separately for the git-based or snapshot builds than I can for the release) at all than a build failure per se?

But there is a different variant of A, wondering why the [snapshot](https://plt.cs.northwestern.edu/snapshots/) build is [also failing](https://plt.cs.northwestern.edu/pkg-build/server/built/fail/drracket-tool-text-lib.txt)?

jeapostrophe:
I believe that these problems are outside of my expertise.

I know that `pkg-build` (and I think `snapshot`) cares about the ring --- https://github.com/racket/pkg-build/blob/master/main.rkt#L1150 --- but I don't think it is safe for me to switch them because `pkg-build` is supposed to be testing the current release, so we actually don't want to switch yet.

Issue #508, opened 2021-09-10
Ctrl+Shift+T to open previously closed tabs 


jestarray:
It seems like racket has adopted `ctrl+T` for new tab, `ctrl+W` for close tab, similar to browsers like ff and chrome, however `ctrl+shif+T` is not bound to opening a recently closed tab? I've been using `ALT+F T` for a while now

rfindler:
This would be a really good one to add. I often miss it as well. The "File|Open Recent" menu's top most item is an approximation of that, but not exactly the same (it will be the most recently visited tab)

sorawee:
FWIW, I created https://docs.racket-lang.org/drracket-restore-workspace/ that implements this functionality (more or less). You need to create a keybinding yourself however.

I believe @Metaxal also implemented something similar with Quickscript.

sorawee:
Oh no, I misread the thread, sorry. Mine restores the entire workspace. The requested feature here restores only one tab.

Metaxal:
It should be mostly straightforward to adapt the open-recent quickscript to
your needs:
https://gist.github.com/Metaxal/0bbd874891442aad78fbd0c08a329cd9

If you do write such a script, it would be great if you could share it on
the wiki:
https://github.com/racket/racket/wiki/Quickscript-Scripts-for-DrRacket

squarebat:
@Metaxal sorry for the delay, couldn't look into this due to exams. I couldn't fix it by upgrading for some reason, but a clean install did the trick! The script works as expected. I will look into how to add that change to the gui repo.

PR #543, opened 2022-01-17
New file menu item - Reopen Closed Tab

squarebat:
As discussed in Issue #508 , this PR implements the functionality for reopening previously closed tabs. It has been added to the file menu after `Open Require Path...` as 'Reopen Closed Tab'. Since `Ctrl-Shift-T` is reserved for 'Spell Check Test' in the edit menu, I gave this the shortcut `Ctrl-*`, which might be a bit unintuitive.

As in [@Metaxal 's quickscript](https://gist.github.com/Metaxal/56d288626574debac24b41fb0232bbe1), the callback `reopen-closed-tab` finds the first file in recently-opened-files that isn't already open and still exists. If no tabs were closed in the current session, it will open files that were active in previous sessions. However, after the recently-opened-files list is exhausted, it will just open new tabs. I thought it would be best for the callback to reside in [unit.rkt](https://github.com/squarebat/drracket/blob/862310dfdc7937fe9bdf352a56e4819b46a9d49f/drracket/drracket/private/unit.rkt) where related methods `create-new-tab` and `open-in-new-tab` are stored.

This has only been manually tested, how should I write the tests for it? I couldn't find an appropriate file in the repository for testing GUI related features. 

I will admit I couldn't look into this issue for a long time, and I am extremely sorry about the delay ;-;

The corresponding PR to add the new string constant reopen-closed-tab - [string-constants/pull/56](https://github.com/racket/string-constants/pull/56)

### DIFF
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -3248,6 +3248,7 @@
               (send tab set-i (- (send tab get-i) 1)))
             (set! tabs (remq tab-to-close tabs))
             (send tabs-panel delete (send tab-to-close get-i))
+            (update-closed-tabs tab)
             (update-menu-bindings)
             (cond
               [(eq? tab-to-close current-tab)
@@ -3273,15 +3274,16 @@
       ;; previous sessions are opened.
       (define/public reopen-closed-tab
         (lambda ()
-          ; Get the list of recently opened files
-          (define recently-opened-files (preferences:get 'framework:recently-opened-files/pos))
+          (define closed-tabs (preferences:get 'framework:recently-closed-tabs))
           (define file-to-open
-            (for/or ([file (in-list recently-opened-files)])
+            (for/or ([file (in-list closed-tabs)])
               (define filename (first file))
               (and (not (find-matching-tab filename))
                    (file-exists? filename)
+                   (set! closed-tabs (cdr closed-tabs))
                    filename)))
           (open-in-new-tab file-to-open)
+          (preferences:set 'framework:recently-closed-tabs closed-tabs)
           #f))
 
       (define/public (get-tab-count) (length tabs))
@@ -3412,6 +3414,12 @@
         (when close-tab-menu-item
           (update-close-tab-menu-item-shortcut close-tab-menu-item))
         (update-close-menu-item-shortcut (file-menu:get-close-item)))
+      
+      (define/private (update-closed-tabs tab)
+        (define tab-filename (send (send tab get-defs) get-filename))
+        (when tab-filename 
+          (define closed-tabs (preferences:get 'framework:recently-closed-tabs))
+          (preferences:set 'framework:recently-closed-tabs (cons (list tab-filename 0 0) closed-tabs))))
       
       (define/private (update-close-tab-menu-item-shortcut item)
         (define just-one? (and (pair? tabs) (null? (cdr tabs))))
@@ -3950,7 +3958,8 @@
                   (when pth (handler:edit-file pth)))])
           (new menu:can-restore-menu-item%
             (label (string-constant reopen-closed-tab))
-            (shortcut #\*)
+            (shortcut #\t)
+            (shortcut-prefix (cons 'shift (get-default-shortcut-prefix)))
             (parent file-menu)
             (callback
             (λ (x y)
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -3274,16 +3274,19 @@
       ;; previous sessions are opened.
       (define/public reopen-closed-tab
         (lambda ()
-          (define closed-tabs (preferences:get 'framework:recently-closed-tabs))
+          (define closed-tabs (preferences:get 'drracket:recently-closed-tabs))
           (define file-to-open
             (for/or ([file (in-list closed-tabs)])
               (define filename (first file))
-              (and (not (find-matching-tab filename))
-                   (file-exists? filename)
+              (and (file-exists? filename)
                    (set! closed-tabs (cdr closed-tabs))
                    filename)))
-          (open-in-new-tab file-to-open)
-          (preferences:set 'framework:recently-closed-tabs closed-tabs)
+          (when file-to-open
+            (let ((tab (find-matching-tab file-to-open)))
+              (if tab
+                (change-to-tab tab)
+                (open-in-new-tab file-to-open))))
+          (preferences:set 'drracket:recently-closed-tabs closed-tabs)
           #f))
 
       (define/public (get-tab-count) (length tabs))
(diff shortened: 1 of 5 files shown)


rfindler:
Thanks for putting some time into this!

I don't think that using `'framework:recently-opened-files/pos` will work. That's not only recently closed files, but it is updated at other times too. I think this needs to be a new preference.

Also, since <menukey>-shift-t is so widely used for this shortcut, I wouldn't mind changing that.

squarebat:
@rfindler on futher exploring, recently-opened-files does behave unexpectedly for this feature. I'll set up a new preference that maintains a list of closed tabs.

rfindler:
Great! The preference will probably need to be initialized and maintained in the racket/gui repo (in the framework, near to where `'framework:recently-opened-files/pos` is maintained).

squarebat:
I have added a new preference `recently-closed-tabs`. This list is updated whenever a tab is closed, and also when a tab is reopened. It works as expected but it is still incomplete with respect to :
- As in [gui/PR257](https://github.com/racket/gui/pull/257), the default test and un/marshall functions for recently-closed-tabs are the same as recently-opened-files. Should we create named procs for the lambdas used?
- In update-closed-tabs, I want to remove redundant tabs and introduce a size limit for the list. This should ideally be done by the add-to-recent method in gui-lib/framework/private/handler.rkt . However it currently handles only recently-opened-files. I tried to create an abstraction (add-to-list) and add another method add-to-closed-tabs. But I couldn't install it correctly and can't access the method outside handler. This might be because I am testing my code on a slightly older version. I'll update and check again.
- The shortcut for Spell Check Test needs to be updated since -shift-t is now reserved for reopen closed tab.

rfindler:
- I think that since the framework doesn't seem to be notified when a tab is closed that maybe it makes the most sense to have this preference managed entirely in drracket itself (so not in the framework). So I'd suggest moving the preference creation and initialization into `drracket/private/main` and just having one pull request.

- I think the preference contain only be files that aren't open. So `reopen-closed-tab` can simply open the first one in the list. When opening a file another way, however, it would need to be removed from the list. This is probably not how browsers work, tho? I'm not sure. Maybe trying an experiment to see what happens when you close a tab and then go visit the same url on your own and then try to reopen it?

- It does make sense to put a limit on the number of files in the preference. A simple way way to do that is to add the new file to the front of list and then check how long the list becomes. If it is too long, then remove things from the end. Something like this:

```
#lang racket

(define (truncate-list l n)
  (define len (length l))
  (cond
    [(<= len n) l]
    [else (drop-right l (- len n))]))

(require rackunit)
(check-equal? (truncate-list '(1 2 3) 10) '(1 2 3))
(check-equal? (truncate-list '(1 2 3) 3) '(1 2 3))
(check-equal? (truncate-list '(1 2 3) 2) '(1 2))
(check-equal? (truncate-list '(1 2 3) 1) '(1))
(check-equal? (truncate-list '(1 2 3 4 5 6) 2) '(1 2))
```  

- The two numbers in the preference (that go along with the filename) are meant to be the insertion position. These can be recovered by calling `get-editor` or `get-definitions-text` and then asking for its `get-start-position` and `get-end-position` and saving them in the preference.

- There is the issue comparing paths to determine if this is "the same file" or not (when opening a file, that file name might need to be removed from the preference). I think probably using `normal-case-path` and `simplify-path` on both paths is the best thing (before then using `equal?` on the result). But it may make sense to look around and see if there are other places where paths are compared and do what they're doing.

squarebat:
- That makes sense. I falsely assumed that all preferences were being handled by the framework.
- Removing opened files might be a bit tricky since there are several ways in which files are being opened and all seem to have different underlying methods. I figured that it could be best dealt with by checking if a file is already opened in reopen-closed-tab. I checked out what the browser does and surprisingly it (mozilla and brave) will open the url even if it is already opened in another. We could do that too.
- There is a size-down method in framework/private/handler.rkt for that. I thought about installing it as a public method but couldn't achieve it. But if the preference doesn't stay in the framework, it may be good to leave framework out entirely. 
- Thanks! I forgot to ask what start and end were being used for and ignored them for the time being.
- The method `recently-opened-files-same-enough-path?` in handler does exactly that. I'll do the same.

rfindler:
I see what you mean about removing items from the list (because they were
opened elsewhere). I think we could achieve that goal if we wanted to (I
can dig around to find the single point of control -- I am sure there is
one or we could make one), but if you think it would be clearer to the
users to not do that, I am on board.

Robby

rfindler:
That's a great point about just moving to the active tab. Indeed, unless we go to some lengths, just opening the file will send them to the tab where it is already open! Great!

squarebat:
>  unless we go to some lengths

Indeed, the only way I was able to open the same file twice was by playing around with `reopen-closed-tabs`
In the latest commit, I have added the preference to drracket/private/main. As discussed, if a closed file has already been opened, `\-shift-t` will redirect to active tab. I have also added code to remove duplicates from list and introduced an upper limit - recently-closed-tabs-max-count.

One problem though. I have retrieved the insertion positions from get-definitions-text, but the values will be lost since `reopen-closed-tabs` calls `(open-in-new-tab filename)` which ultimately calls `create-new-tab` where a new text object is created. Updating that new text object with the saved positions is fairly straightforward, but it'll require changing the create-new-tab method and adding default value parameters to it. Shall I go ahead with that?

rfindler:
I think adding a keyword, optional argument that specifies the position to `create-new-tab` and `open-in-new-tab` is great. (The docs would need to be updated too, tho.)

I think that `'drracket:recently-closed-tabs-max-count` can default to a larger number. The 50 in the other one is because they get put into a menu somewhere. But these won't be.

Because the definition of `truncate-list` is in a class, it is going to be a private field (i.e. per-object storage), so probably best to change it to `define/private` so that it will be a method (no extra storage in the object it is just once in the class).

Probably good to remove the shortcut for the other item -- or is that not in DrR? I'm forgetting where it is.

And there eventually needs to be an update to `HISTORY.txt` and an update to the documentation because there is a new menu item.

drracket/drracket/private/unit.rkt
```diff
@@ -3075,7 +3075,7 @@
       ;; create-new-tab : -> void
       ;; creates a new tab and updates the GUI for that new tab
       (define/public create-new-tab
-        (lambda ([filename #f])
+        (lambda ([filename #f] [start-pos 0] [end-pos 'same])
```
rfindler, line 3078: I think keyword arguments here is better, eg:
```
(lambda ([filename #f] #:start-pos [start-pos 0] #:end-pos [end-pos 'same])
```

drracket/drracket/private/unit.rkt
```diff
@@ -3266,8 +3267,8 @@
            #t]
           [else #f]))
       
-      (define/public (open-in-new-tab filename)
-        (create-new-tab filename))
+      (define/public (open-in-new-tab filename [start-pos 0] [end-pos 0])
+        (create-new-tab filename start-pos end-pos))
```
rfindler, line 3271: ditto here for the keywords.

squarebat:
- Updated the args. They do make more sense as kwargs
- Should History.txt be updated to reflect the addition of the keyword args?

squarebat:
Are there any updates on this?

drracket/drracket/private/unit.rkt
```diff
@@ -3265,9 +3267,31 @@
...
+      ;; Opens previously closed tabs. If no tabs were closed in current session, files from 
+      ;; previous sessions are opened.
+      (define/public reopen-closed-tab
+        (lambda ()
+          (define closed-tabs (preferences:get 'drracket:recently-closed-tabs))
+          (define file-to-open
+            (for/or ([file (in-list closed-tabs)])
+              (define filename (first file))
+              (and (file-exists? filename)
+                   (set! closed-tabs (cdr closed-tabs))
```
rfindler, line 3283: Does this work properly? Maybe better to remove `file-to-open` from `closed-tabs` after this loop completes? And if any files were skipped because they don't exist anymore, those should probably be removed from the preference too.

drracket/scribblings/tools/unit.scrbl
```diff
@@ -619,15 +619,23 @@ Returns the currently active tab.
...
 }
 
-@defmethod[(open-in-new-tab [filename (or/c path-string? #f)]) void?]{
+@defmethod[(open-in-new-tab [filename (or/c path-string? #f)]
+                            [#:start-pos start-pos exact-nonnegative-integer? 0]
+                            [#:end-pos end-pos (or/c exact-nonnegative-integer? 'same) 'same]) 
+                            void?]{
   Opens a new tab in this frame. If @racket[filename] is a @racket[path-string?],
-  It loads that file in the definitions window of the new tab.
+  It loads that file in the definitions window of the new tab. If @racket[start-pos] and
```
rfindler, line 627: Lowercase "i" in "it".

rfindler:
Sorry for the long delay here @squarebat . I think that it is basically ready to go. I can address the two comments I put in and merge it, if you want?

squarebat:
Sure!

PR #589, opened 2022-09-22
Make pressing Kill button call kill-evaluation

### LINKED ISSUE #588: "Kill" button is misleading
After we click the "Stop" button, the label is changed to "Kill". This "Kill" seems to have a different meaning from Cmd+K (Force the Program to Quit), however. I'm not totally sure how they are different precisely, but I do know that, for example, clicking the "Kill" button won't invoke the `kill-evaluation` method, but Cmd+K will.

The different behavior is confusing and misleading. They should have the same behavior. Alternatively, there should be three different labels for these three different actions.

sorawee:
Fixes #588

### DIFF
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -1329,10 +1329,8 @@
                  (not custodian-to-kill))
              (bell)]
             [do-kill?
-             (cond
-               ;; this transitively shuts down custodian-to-kill too
-               [ints (send ints kill-evaluation)]
-               [else (custodian-shutdown-all custodian-to-kill)])
+             ;; this transitively shuts down custodian-to-kill too
+             (send ints kill-evaluation)
              (set! do-kill? #f)
              (send (get-frame) update-kill-button-label)]
             [else
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -1329,8 +1329,10 @@
                  (not custodian-to-kill))
              (bell)]
             [do-kill?
-             (when custodian-to-kill
-               (custodian-shutdown-all custodian-to-kill))
+             (cond
+               ;; this transitively shuts down custodian-to-kill too
+               [ints (send ints kill-evaluation)]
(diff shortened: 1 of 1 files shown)


drracket/drracket/private/unit.rkt
```diff
@@ -1329,8 +1329,10 @@
                  (not custodian-to-kill))
              (bell)]
             [do-kill?
-             (when custodian-to-kill
-               (custodian-shutdown-all custodian-to-kill))
+             (cond
+               ;; this transitively shuts down custodian-to-kill too
+               [ints (send ints kill-evaluation)]
+               [else (custodian-shutdown-all custodian-to-kill)])
```
sorawee, line 1335: I think the else clause is a dead code, but I put it there just to be safe and consistent with previous code.

rfindler: I think it is okay to remove the else clause. Generally speaking, there aren't guards on uses of `ints` -- it should only be unset as the objects are all be created and the callback shouldn't happen until after they are all set up (because the GUI objects won't be visible until after things are all set up).

Well, I'm not 100% sure of that, but I am fairly confident that we'd fix it elsewhere if `ints` could be `#f` here.

sorawee: Done.

sorawee:
@rfindler As you suggested, I investigated whether `kill-evaluation` was added after `break-callback`, and the answer is that I do not know. They were both added in the same commit https://github.com/racket/drracket/commit/b1ca4cec644125f01bb174a4d8d047d7d056ffa3, but it appears there might be further history than that, which I am unable to follow.

That being said, both `kill-evaluation` and `break-callback` (in the kill mode) used to simply shutdown the custodian. Clearing ports was added in https://github.com/racket/drracket/commit/fb702bd352f98456bc69dbb789008ef8697fb3ea, but this only applies to `kill-evaluation` and not `break-callback`'s kill mode.

rfindler:
Thank you. This looks like an oversight in https://github.com/racket/drracket/commit/fb702bd352f98456bc69dbb789008ef8697fb3ea, then!

PR #592, opened 2022-10-10
Allow adding Debug button to #lang languages

mikesperber:
Fixes https://github.com/racket/htdp/issues/177

### DIFF
--- a/drracket/drracket/private/insulated-read-language.rkt
+++ b/drracket/drracket/private/insulated-read-language.rkt
@@ -24,6 +24,7 @@ Will not work with the definitions text surrogate interposition that
          syntax-color/racket-lexer
          syntax-color/lexer-contract
          drracket/syncheck-drracket-button
+         gui-debugger/debug-tool-button
          (for-syntax racket/base))
 
 (define recognized-read-language-symbol/c
@@ -310,7 +311,8 @@ Will not work with the definitions text surrogate interposition that
     ;; is in the invocation of the local-member-name in the callback
     ;; that passes the drracket frame there that requires this to be
     ;; in the irl namespace.
-    (namespace-attach-module trusted-namespace 'drracket/syncheck-drracket-button))
+    (namespace-attach-module trusted-namespace 'drracket/syncheck-drracket-button)
+    (namespace-attach-module trusted-namespace 'gui-debugger/debug-tool-button))
   ns)
 
 (define (pick-new-language text all-languages module-language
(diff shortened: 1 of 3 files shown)


soegaard:
@rfindler This PR seems to be forgotten?

rfindler:
Thank you for the reminder @soegaard ! I did somehow miss this one. 

@mikesperber  I think that maybe changing [something around here](https://github.com/racket/htdp/blob/master/htdp-lib/htdp/bsl/reader.rkt#L28-L37) based on [these docs](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28part._.Opting_.Out_of_.Standard_.Toolbar_.Buttons%29) is a better way to go. Does that not work?

mikesperber:
> @mikesperber I think that maybe changing [something around here](https://github.com/racket/htdp/blob/master/htdp-lib/htdp/bsl/reader.rkt#L28-L37) based on [these docs](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28part._.Opting_.Out_of_.Standard_.Toolbar_.Buttons%29) is a better way to go. Does that not work?

Yes: But that has this as a prerequisite.

rfindler:
> > @mikesperber I think that maybe changing [something around here](https://github.com/racket/htdp/blob/master/htdp-lib/htdp/bsl/reader.rkt#L28-L37) based on [these docs](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28part._.Opting_.Out_of_.Standard_.Toolbar_.Buttons%29) is a better way to go. Does that not work?
> 
> Yes: But that has this as a prerequisite.

When I change the quoted code to have these two lines:

```
    [(drscheme:opt-out-toolbar-buttons)
     '(macro-stepper)]
```

(instead of returning `#f` from that case), then I see a debug button and clicking on it seems to turn on the debugger. 

Can you please elaborate on what you tried and how it goes wrong that led you to want to make modifications to drracket itself here?

mikesperber:
@rfindler Ah, you mean the bottom line of that snippet, not the top lines. (The docs you quote are almost impossible to find if you don't know about them beforehand.)

I was following the example of the syntax-check and stepper buttons - the top lines of the snippet.

Shouldn't those be done the same way?

rfindler:
I'm still seeing two "Check Syntax" buttons in the teaching languages ASL with the current commit.

Screenshot 2024-09-12 at 3 56 48 PM

shhyou:
Superseded by @mikesperber's racket/htdp#222.

PR #611, opened 2023-03-03
Run Resyntax on drracket/private

jackfirth:
This pull request runs Resyntax on all the reasonably-sized files in the `drracket/private` collection. It omits changes made to any of the huge files in this collection in order to make this pull request easier to review.


samth:
Would it be possible for resyntax to print out a list of all the changes it made (something like "let->define, foo.rkt line 12")?

jackfirth:
Yes, it currently does that in the command line. Unfortunately, that doesn't help much in this case because I only committed a subset of the changes. The full change report was this:

```
resyntax: --- summary ---

  Fixed 670 issues in 54 files.

  * Fixed 419 occurences of let-to-define
  * Fixed 55 occurences of cond-let-to-cond-define
  * Fixed 37 occurences of if-let-to-cond
  * Fixed 21 occurences of for-each-to-for
  * Fixed 21 occurences of define-lambda-to-define
  * Fixed 18 occurences of if-else-false-to-and
  * Fixed 12 occurences of if-begin-to-cond
  * Fixed 10 occurences of hash-ref-with-constant-lambda-to-hash-ref-without-lambda
  * Fixed 9 occurences of nested-for-to-for*
  * Fixed 8 occurences of hash-set!-ref-to-hash-update!
  * Fixed 7 occurences of de-morgan-or-to-and
  * Fixed 6 occurences of cond-else-cond-to-cond
  * Fixed 6 occurences of ormap-to-for/or
  * Fixed 6 occurences of nested-if-to-cond
  * Fixed 6 occurences of single-argument-plus-to-identity
  * Fixed 5 occurences of cond-else-if-to-cond
  * Fixed 5 occurences of de-morgan-and-to-or
  * Fixed 3 occurences of andmap-to-for/and
  * Fixed 3 occurences of sort-with-keyed-comparator-to-sort-by-key
  * Fixed 2 occurences of always-throwing-cond-to-when
  * Fixed 2 occurences of flat-contract-migration
  * Fixed 2 occurences of inverted-when
  * Fixed 2 occurences of define-case-lambda-to-define
  * Fixed 1 occurence of syntax-rearm-migration
  * Fixed 1 occurence of syntax-disarm-migration
  * Fixed 1 occurence of list->vector-for/list-to-for/vector
  * Fixed 1 occurence of equal-null-list-to-null-predicate
  * Fixed 1 occurence of let-values-then-call-to-call-with-values
```

One way we could surface this information better is by making Resyntax create multiple commits and group each change category into a single commit.

drracket/drracket/private/drracket-normal.rkt
```diff
@@ -298,23 +300,24 @@
...
+                    (λ (x y)
+                      (break-thread to-break)))
+                   (parent f))]
+           [b (new button% 
+                   (label "Break All Threads")
+                   (callback
+                    (λ (x y)
+                      ((dynamic-require 'drracket/private/key 'break-threads))))
+                   (parent f))])
+      (send f show #t))))
```
rfindler, line 320: This could be turned into a `define` for `f` and then two things that just don't bind the buttons because `b` isn't actually referenced?

samth:
> One way we could surface this information better is by making Resyntax create multiple commits and group each change category into a single commit.

This sounds very nice.

drracket/drracket/private/frame.rkt
```diff
@@ -670,16 +670,16 @@
...
-                             (floor (* 100 (/ total size)))))) 
-                    (let ([s (read-string 1024 port)]) 
+                             (floor (* 100 (/ total size))))))
+                    (define s (read-string 1024 port))
+                    (unless (eof-object? s) 
                       (unless (eof-object? s) 
-                        (unless (eof-object? s) 
-                          (display s) 
-                          (loop (+ total (string-length s))))))))
+                        (display s)
```
rfindler, line 681: doubled `unless`! Guess that won't be a common error :)

drracket/drracket/private/frame.rkt
```diff
@@ -872,10 +873,9 @@
     (define sorted-items (sort items cmp))
     
     (define (different-slots? item-key next-item-key)
-      (or (not (= (quotient item-key 100)
-                  (quotient next-item-key 100)))
-          (not (= (sgn item-key)
-                  (sgn next-item-key)))))
+      (not (and (= (quotient item-key 100)
+                   (quotient next-item-key 100)) (= (sgn item-key)
+                                                    (sgn next-item-key)))))
```
rfindler, line 878: I don't think I like this one. I can see why it was done and it does seem like a good idea in some cases, but I think that the slots are different in two situations; either the one thingy is different or the other thingy was different. Renaming the function to `same-slots?` and putting a `not` at the callsite works, however!

drracket/drracket/private/module-language-tools.rkt
```diff
@@ -331,8 +331,7 @@
       (define hash-lang-comment-end-position #f)
       (define/private (set-hash-lang-comment-end+last-location _hash-lang-last-location
                                                                _hash-lang-comment-end-position)
-        (unless (or (and (not _hash-lang-last-location)
-                         (not _hash-lang-comment-end-position))
+        (unless (or (not (or _hash-lang-last-location _hash-lang-comment-end-position))
```
rfindler, line 334: This one doesn't seem helpful to me.

drracket/drracket/private/module-language-tools.rkt
```diff
@@ -662,9 +661,9 @@
       
       (define extra-default-filters '())
       (define default-extension "")
-      (define indentation-function (λ (x y) #f))
+      (define (indentation-function x y) #f)
       (define/public (get-indentation-function) indentation-function)
-      (define range-indentation-function (λ (x y z) #f))
+      (define (range-indentation-function x y z) #f)
```
rfindler, line 666: I think this refactoring should not be done (indeed it should go the other way) when the function ends up being a field declaration in a class. Thsi helps to make it clear to the programmer that they're using per-instance space, not per-class space.

drracket/drracket/private/stack-checkpoint.rkt
```diff
@@ -121,16 +121,16 @@
   (define test-suite-filename (syntax-source #'here))
 
   (let ()
-    (define (a x) (+ (f x)))
+    (define (a x) (f x))
```
rfindler, line 124: These changes are wrong. This breaks the program. (All the ones in this file).

drracket/drracket/private/syncheck/blueboxes-gui.rkt
```diff
@@ -317,7 +317,7 @@
       (set! to-invalidate (currently-drawn-bluebox-region))
       (maybe-invalidate))
     
-    (define pref-changed-callback (λ (x y) (invalidate-blue-box-region)))
+    (define (pref-changed-callback x y) (invalidate-blue-box-region))
```
rfindler, line 320: This is another one where I wrote it that way on purpose to emphasize that we've not got a method so that we're holding onto a reference to the object to keep the callback registration alive.

drracket/drracket/private/syncheck/blueboxes-gui.rkt
```diff
@@ -524,9 +524,10 @@
                                       [i (in-naturals)])
                              (define bbs
                                (fetch-blueboxes-strs meth-tag #:blueboxes-cache the-blueboxes-cache))
-                             (if (zero? i)
-                                 (or bbs '())
-                                 (if bbs (cdr bbs) '()))))
+                             (cond
+                               [(zero? i) (or bbs '())]
+                               [bbs (cdr bbs)]
+                               [else '()])))
```
rfindler, line 530: This is a minor comment but this change feels strange. It is like there should be two cases, when `i` is zero or not. And the `cdr`ing of `bbs` isn't really a third case; it is a subcase situation when we're not on the first iteration of the loop. Basically, it is keeping the first method header thingy, but not the other ones. Is there a better way to express that?

drracket/drracket/private/tooltip.rkt
```diff
@@ -87,7 +87,7 @@
       (send parent reflow-container))
 
     ;; private field for gc reasons
-    (define font-size-callback (λ (p v) (set! tooltip-font #f) (update-size)))
+    (define (font-size-callback p v) (set! tooltip-font #f) (update-size))
```
rfindler, line 90: This is another situation where I wanted to emphasize that this is per-object storage (so we keep the pref callback alive as long as the object is alive but they can both be gc'd afterwards).

rfindler:
> > One way we could surface this information better is by making Resyntax create multiple commits and group each change category into a single commit.
> 
> This sounds very nice.

I agree!

drracket/drracket/private/drracket-normal.rkt
jackfirth: I noticed that one too. Resyntax doesn't currently detect this case, but I opened https://github.com/jackfirth/resyntax/issues/185 for the feature request.

drracket/drracket/private/frame.rkt
jackfirth: Yeah I can't imagine that one happening more than once or twice. Sometimes you just gotta be sure :p

drracket/drracket/private/frame.rkt
jackfirth: The rules related to De Morgan's laws have felt a little confusing to me in some cases too. I think they're not worth keeping.

drracket/drracket/private/module-language-tools.rkt
jackfirth: De Morgan strikes again.

drracket/drracket/private/module-language-tools.rkt
jackfirth: Hmm, I could get behind that. In this case it's also weird because of how the functions are named: `(define indentation-function (lambda (x y) #f)))` looks fine and `(define (indentation x y) #f)` looks fine, but `(define (indentation-function x y) #f)` looks weird.

drracket/drracket/private/stack-checkpoint.rkt
jackfirth: Interesting. Are these functions trying to test the behavior of the stack checkpoints in a way that makes whether they're calling each other in tail position important?

drracket/drracket/private/syncheck/blueboxes-gui.rkt
jackfirth: Perhaps a named helper function or a comment? Neither the nested `if` structure nor the `cond` structure conveys that information to me on its own.

drracket/drracket/private/tooltip.rkt
jackfirth, line 90: Another callback that shouldn't be delambdified.

drracket/drracket/private/syncheck/blueboxes-gui.rkt
mfelleisen: In class (and personally), I'd write this as 

```
(cond 
  [(zero? i) (or bbs '())]
  [else (if bbs (car bbs) '())])
```

to indicate with `cond` that I am dispatching on Naturals and that the nested condition is for iterations 2+.

drracket/drracket/private/frame.rkt
rfindler: Does it make sense to have some kind of "boolean normalization" rules that search for the smallest equivalent expression and then, if it is especially smaller than the current one, suggest it?

rfindler: That is, I have found that thinking at that level (boolean identities) has been helpful even if I don't know how it is going to turn out in the end; seeing how it turns out often helps me decide if I like it or not.

drracket/drracket/private/module-language-tools.rkt
rfindler: Great!

This also reminds me: there are definitely some bugs out there where people wrote `(define (f x ...) body)` inside a class an didn't realize that they are doing something bad and they should have written `(define/private (f x ...) body)`. If resyntax offers the "make a lambda to draw the eye"-type refactoring they might realize "oh! I wanted `define/private` instead". Or resyntax can even somehow tell them if it can comment on its suggestions somehow.

drracket/drracket/private/stack-checkpoint.rkt
rfindler: Yes, exactly.

rfindler: I am happy to change the test case so that resyntax knows "this is a non-tail call on purpose". Maybe just using `list` or something would be good. Or maybe defining the identity function and calling it, giving it a suggestive name?

drracket/drracket/private/syncheck/blueboxes-gui.rkt
rfindler: Actually, this code is super confusing, isn't it!

drracket/drracket/private/frame.rkt
jackfirth: I think I'd rather not even try to solve it with automation. There's bigger fish to fry, and making sure it never makes things worse seems very difficult. The ideal suggestions for Resyntax to make are those that most people consider clearly and significantly better.

drracket/drracket/private/stack-checkpoint.rkt
jackfirth: I think defining the identity function with a suggestive name would be the ideal approach. Something like this:

```scheme
(define (ensure-non-tail-call v) v)

(define (a x) (ensure-non-tail-call (f x)))
```

drracket/drracket/private/syncheck/blueboxes-gui.rkt
rfindler: I rewrote this in a way that @jackfirth probably imagines.

drracket/drracket/private/stack-checkpoint.rkt
rfindler: Sounds good. I've made that change.

drracket/drracket/private/module-language-tools.rkt
jackfirth: Perhaps Resyntax could rewrite `(define (f x ...) body ...)` to `(define/private (f x ...) body ...)` whenever it's within a class and `f` is never mutated? Would that always be correct?

rfindler:
No, not always. Some of these examples would go wrong actually (for GC
reasons).

But maybe if it only appears in the function position of applications (and
never in other positions) the it should be a good suggestion. The things
that are happening here are definitely very special cases and if they
aren't being used like that, we can easily make a change to block
resyntax's suggestion by adding that kind of reference.

It would be really great if resyntax noticed something that should be a
private method and os currently a private field bound to a procedure!

Robby

jackfirth:
@rfindler Could you file a feature request for that? https://github.com/jackfirth/resyntax

drracket/drracket/private/stick-figures.rkt
```diff
@@ -101,67 +101,63 @@
...
-    (let-values ([(min-x min-y) (get-max/min-x/y min points)])
-      (map (λ (x) (list (car x) 
-                        (+ (- (list-ref x 1) min-x))
-                        (+ (- (list-ref x 2) min-y))))
-           points)))
+  (define-values (min-x min-y) (get-max/min-x/y min points))
+  (map (λ (x) (list (car x) 
+                    (- (list-ref x 1) min-x)
+                    (- (list-ref x 2) min-y)))
+       points))
```
cloudrac3r, line 160: `map` might be better as `for/list` here, since it avoids the dangling `points))` line. Could be a good heuristic/test case for Resyntax?

```
(for/list ([x points])
  (list (car x)
        (- (list-ref x 1) min-x)
        (- (list-ref x 2) min-y)))

jackfirth: Yeah, I think that's worth doing. It could be similar to the current `for-each` to `for` rule, which only fires when `for-each` uses a multi-line lambda. Can you create an issue in the Resyntax repository for this?

cloudrac3r: jackfirth/resyntax#193

jackfirth:
I'm closing this one in favor of #612.

PR #617, opened 2023-06-07
add syncheck:import-or-export-prefix-ranges

rfindler:
Not sure if this is immediately (or ever?) useful, but it was easy to make and it seems like it fits here, so I made it.

Related to racket/racket#4649

### DIFF
--- a/drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
+++ b/drracket-tool-doc/scribblings/drracket-tools/drracket-tools.scrbl
@@ -472,6 +472,25 @@ in order to make the results be platform independent.
     This method is listed only for backwards compatibility. It is not called
     by Check Syntax anymore.
   }
+
+
+ @defmethod[(syncheck:import-or-export-prefix-ranges
+             [id identifier?]
+             [ranges (listof (vector/c natural?
+                                       natural?
+                                       identifier?
+                                       #:immutable #t))])
+            void?]{
+  This method is called to indicate the locations of prefixes of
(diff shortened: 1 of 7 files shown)


rfindler:
Oh! Okay! Thanks, Greg.

greghendershott:
If I should make a fresh issue for this, please let me know, but:

A potential to-do would be for check-syntax to add arrows for prefixes described by this new syntax property.

In other words, generalize how it handles the single prefix case, indicated by the low-level `prefix` clause in `#%require`, to one or more prefixes.

So for things like `(prefix-in A: (prefix-in B: mod))` it would do more `add-arrow-*`s and more `add-prefixed-require-reference`s, for both `A:` and `B:`, whereas today it does none.

[The only reason I need the syntax property, for imports^1, is to draw more arrows that check-syntax currently doesn't draw for me in the first place.^2]

Is that something you'd be willing to do? (I took an initial look in `traversals.rkt`. I feel like you'd be able to do this more quickly -- and to your greater satisfaction -- than me attempting it. And I have plenty else on my plate right now. But if you don't have time I can attempt this later and do a PR.)

---

^1: For prefixed _exports_ it's a different, bigger story, involving potentially both this property and sub-range-binders (e.g. `(prefix-out foo- (struct-out bar))`). But I think that's much lower-priority; I have it working in pdb, and the value to end users visually in e.g. DrRacket seems much less.
 
^2: There are some other reasons I draw more arrows, but the prefix arrows are one subset.

rfindler:
I can give this a try, yes and thanks for pointing out the lack!

PR #620, opened 2023-06-22
add `drracket:define-popup` configuration to `#lang`-based `get-info`

mflatt:
This change allows Rhombus to configure DrRacket so that the `(define ...)` menu becomes a `def` menu that jumps to Rhombus definitions.

Besides allowing `#lang`-based configuration, the change refines how the definition popup works:

 * Instead of looking for Scheme-style comments, recognize text that should not match based on its syntax-coloring category, ignoring would-be matches in comments or in string literals.

 * Allow the configuration to specify case-sensitive matching and delimited matching, where the definition of "delimited" is derived from the expression-navigation configuration.

### DIFF
--- a/drracket-plugin-lib/drracket/private/drsig.rkt
+++ b/drracket-plugin-lib/drracket/private/drsig.rkt
@@ -149,7 +149,9 @@
    (struct online-expansion-handler (mod-path id local-handler monitor?))
    get-online-expansion-handlers
    no-more-online-expansion-handlers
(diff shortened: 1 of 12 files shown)


mflatt:
Updated with test repairs, thanks to Robby.

rfindler:
I've squashed these commits and pushed to master: https://github.com/racket/drracket/commit/fc48296eef63dee1f301a3d1da54456e97283557

PR #621, opened 2023-06-28
generalize `drracket:define-popup` to callbacks

mflatt:
Predictably, having just a pair of symbol options is not configurable enough for the popup "define" menu. Remove the option symbols and instead support callback functions for the definition-finding and name-extracting tasks. Each callback receives the default implementation, in case it finds that helpful, and the former option symbols have turned into optional keyword arguments for the default definition-finding implementation.

Rhombus branch that uses the new protocol: https://github.com/mflatt/rhombus-prototype/tree/define-popup2, implementation in https://github.com/mflatt/rhombus-prototype/blob/define-popup2/rhombus/private/define-popup.rkt.

### DIFF
--- a/drracket/drracket/private/get-defs.rkt
+++ b/drracket/drracket/private/get-defs.rkt
@@ -15,7 +15,7 @@
 (define-struct defn (indent name start-pos end-pos)
   #:mutable #:transparent)
 
-(struct define-popup-info (prefix long-name short-name case-sensitive? delimited?) #:transparent)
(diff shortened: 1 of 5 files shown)


rfindler:
Not having done any testing myself, I think this looks okay.

rfindler:
I've pushed this to master -- one note: the guards added here are incorrect; they should be thunks, not just the default values.

Thanks!

PR #623, opened 2023-07-02
add pubment after-create-new-drracket-frame and after-create-new-tab

Metaxal:
New pubment methods to be used in next PRs by Quickscript.

### DIFF
--- a/drracket/drracket/private/interface.rkt
+++ b/drracket/drracket/private/interface.rkt
@@ -36,7 +36,9 @@ remain the same for tools that use them.
     get-definitions-canvas
     get-button-panel
     execute-callback
+    after-create-new-drracket-frame
     get-current-tab
+    after-create-new-tab
     open-in-new-tab
     create-new-tab
     close-current-tab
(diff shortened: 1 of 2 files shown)


Metaxal:
Currently the `queue-callback` for `create-new-drracket-frame` is in quickscript, but I wonder whether it should be in the pubment method instead.

rfindler:
I think we need documentation too.

Why is there a guard on `filename` for the call to `after-create-new-tab`?

Metaxal:
Yes, docs, will do. 

For filename, I wanted to call the hook only when an empty new tab is created. For loading file, I think augmenting `on-file-load` in editor<%> is more appropriate.

That said, the pubment method could well accept all the same arguments, and QS could do the guarding itself. Would you prefer this option?

rfindler:
I don't have a strong opinion on any specific timing of the callbacks, actually, just as long as the entire set makes some sense together is all. I guess the docs will clarify it all, then!

Metaxal:
As the message says, I removed the `filename` guard and added the docs. LMKWYT.

Issue #634, opened 2023-08-26
Proposal: New info key `drracket:comments`


greghendershott:
Things like DrRacket and Emacs offer commands to comment and un-comment a region.

Non-s-expression langs should be able to supply information for these commands, via a new info key named something like "`drracket:comments`".

Open to improvements, but initially I'll suggest the value is a list:

```
(list comment-start comment-end comment-padding)
```

where those three values as identical to three configuration variables in Emacs:

>    When a comment command makes a new comment, it inserts the value of
> ‘comment-start’ as an opening comment delimiter.  It also inserts the
> value of ‘comment-end’ after point, as a closing comment delimiter.  For
> example, in Lisp mode, ‘comment-start’ is ‘";"’ and ‘comment-end’ is
> ‘""’ (the empty string).  In C mode, ‘comment-start’ is ‘"/* "’ and
> ‘comment-end’ is ‘" */"’.
> 
>    The variable ‘comment-padding’ specifies a string that the commenting
> commands should insert between the comment delimiter(s) and the comment
> text.  The default, ‘" "’, specifies a single space.  Alternatively, the
> value can be a number, which specifies that number of spaces, or ‘nil’,
> which means no spaces at all.

Examples:

- s-expression langs (default): `(";;" "" " ")`
- scribble: `("@;" "" " ")`
- rhombus: `("//" "" " ")`

I'd suggest that when the key isn't supplied, it's OK to default to s-expression as the default. (Justification: There exist many more s-expression langs.)

---

Whatever the final form, I think the to-do list would be:

- [ ] Document at https://docs.racket-lang.org/tools/lang-languages-customization.html
- [ ] Add the info key to at least a few langs like:
   - [ ] scribble
   - [ ] rhombus
- [ ] Revise DrRacket's (un)comment commands to use this.

If agreed/confirmed, I would probably be able to make PRs easily for all but the last. (I'm not familiar with that code. I'm not sure about details like whether a "Racket" menu is where these should continue to live.)

---

Background: https://github.com/greghendershott/racket-mode/pull/661#issuecomment-1693769408

sorawee:
In C, I expect the configuration to be `("/*" "*/" "")`. Now, If I make a selection that spans multiple lines and hit comment, it probably should wrap `/*`and `*/` around the selection.

But in Rhombus, if I make a selection that spans multiple lines and hit comment, it probably should add `# ` on every line.

How do we distinguish these two scenarios?

rfindler:
[...]

PR #635, opened 2023-08-30
Document drracket:comment-delimiters

greghendershott:
The documentation portion of #634.

Note that this commit does /not/ add to the documented list of keys read by Dr Racket, because this commit does not add the behavior of Dr Dracket doing that and using the comment styles to drive its (un)comment commands.

### DIFF
--- a/drracket/info.rkt
+++ b/drracket/info.rkt
@@ -67,7 +67,7 @@
 
 (define pkg-authors '(robby))
 
-(define version "1.14")
+(define version "1.15")
 
 (define license
   '(Apache-2.0 OR MIT))
--- a/drracket/scribblings/tools/lang-tools.scrbl
+++ b/drracket/scribblings/tools/lang-tools.scrbl
@@ -186,7 +186,7 @@ string used to end/close a comment, and @racket[_padding] goes
 after/before.
 
    Racket example: @racket['(region "#|" "|#" "  " " ")].
-   C++ example: @racket['(region "/*" "*/" " *" " ")].}
+   C++ example: @racket['(region "/*" " *" "*/" " ")].}
 
  ]
 
--- a/drracket/scribblings/tools/lang-tools.scrbl
+++ b/drracket/scribblings/tools/lang-tools.scrbl
@@ -183,6 +183,12 @@ strings:
 
  The default value is @racket[(list (list ";;" "" " ") (list "#|" "|#" " "))].
 
+ When the list has multiple styles, some tools may present them for
+the user to pick one. Other tools may default to using the first style
+in the list (the user will configure a preference by others means).
+Therefore when a language supports multiple comment styles, it should
+list the most popular or preferred style first.
+
  @history[#:added "1.15"]
 }
 
--- a/drracket/scribblings/tools/lang-tools.scrbl
+++ b/drracket/scribblings/tools/lang-tools.scrbl
@@ -167,21 +167,36 @@ The default value is @racket[(list #\" #\|)].
 
  When a language's @racket[_get-info] procedure responds to
 @racket['drracket:comment-delimiters], it is expected to return a list
-of comment styles. Each comment style is expressed as a list of three
-strings:
+of comment styles. Each comment style is expressed as one of:
 
  @itemlist[
 
-  @item{The characters used to start/open a comment.}
+  @item{@racket[(list 'line start padding)], where @racket[_start]
+   is a string used to start a comment that is teriminated by the
+   end of a line, and @racket[_padding] goes after that.
+   Lisp example: @racket['(line ";;" " ")].
+   C++ example: @racket['(line "//" " ")].}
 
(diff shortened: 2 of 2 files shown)


rfindler:
I think it probably makes sense to clarify how newlines are going to be treated. Right now it reads like the empty string is a special case for newline handling. Is that right? I am realizing now that I may have misunderstood a detail from the previous discussion. If we're expecting tools to (for now at least) always put these in at the start and end of lines then we would not expect newlines to be inserted but instead the closing one to be inserted at the end of each line, right before the newline. If that's the case, then it probably helps to explain that in the docs so people understand better what characters to put in there. If it isn't the case (ie these are more general) then we might explain how they are going to be used but point out that they might be used in a more general way in the future.

Also, do we want to include some examples?

greghendershott:
> I think it probably makes sense to clarify how newlines are going to be treated. Right now it reads like the empty string is a special case for newline handling. Is that right?

Yes.

> I am realizing now that I may have misunderstood a detail from the previous discussion. If we're expecting tools to (for now at least) always put these in at the start and end of lines then we would not expect newlines to be inserted but instead the closing one to be inserted at the end of each line, right before the newline. If that's the case, then it probably helps to explain that in the docs so people understand better what characters to put in there. If it isn't the case (ie these are more general) then we might explain how they are going to be used but point out that they might be used in a more general way in the future.

- When the comment-end delimiter is `""`, meaning newline, then the delimiters are for **line** comments.

  The existing, line-oriented DrRacket comment and un-comment commands should still work fine, just substituting the comment-start+comment-padding string for the hard-coded `#\;` character (IIUC).

   - Note: When the user selects a region that isn't entire lines, DrRacket rounds it up to entire lines. That's fine! Another possibility, which Emacs does, is to split the line(s) so that line-comment delimiters can still be used. e.g. With `2` selected here:
 
     ```
     (if 1 2 3)
     ```
     the Emacs comment command will do:
     ```
     (if 1 ;; 2
      3)
     ```
     But this proposal is not saying that you need to, or even ought to, do this.

- When the comment-end delimiter is _not_ `""`, then the delimiters are for **region** comments. 

  The existing, full-line-oriented DrRacket comment and un-comment commands should still work fine, just substituting the comment-start+comment-padding string for the hardcoded `#\;` character (IIUC), and adding the comment-padding+comment-end string just before newline.

  - Making DrRacket work on non-full-lines, and with the region comment delimiters, is IIUC trickier because you have to deal with nested comments? Honestly I haven't thought this through deeply, I'm relying on Emacs comment commands and just giving it the delimiter values.

> Also, do we want to include some examples?

I thought the default value was a good enough example, but I can add more.

rfindler:
Okay, this is all making sense to me. Thanks for the slow (repeated!) explanation. Also, it looks like Sublime Text also rounds-up when doing line comments. (I wonder if Emacs has a configuration parameter to get that behavior?)

Also, would it be clearer if we used two element lists to mean line comments and three element lists to be region comments? That way the empty string might be less confusing to the documentation reader?

sorawee:
Would it be a good idea to tag the "style" (inline vs region) explicitly in the list? I'm thinking about forward compatibility where we might want to extend the list even longer to accommodate e.g. whether nested region comment is supported, and if we use the list length as a way to distinguish style, it will be more difficult to do this extension.

greghendershott:
@rfindler @sorawee I think that sounds good. To make sure I understand, were you thinking something like:

```
(or/c (list 'line start padding)
      (list 'region start end padding))
```

?

And, I could add a bit in the doc to summarize some of what's in the thread above about line vs. region comments, as well as tool choices?

rfindler:
yes, this sounds great;

> And, I could add a bit in the doc to summarize some of what's in the thread above about line vs. region comments, as well as tool choices?

Especially this part!

greghendershott:
I just pushed a commit trying to incorporate this feedback. Please let me know if this seems better, and any corrections/suggestions.

rfindler:
It would be nice to include contracts to make things a bit more precise. Maybe where it says "list of comment styles" it can instead have the contract there? (That's kinds of far away from where they are discussed; maybe repetition is okay or maybe it is better to do a different change.)

I was inspired to think of this by wondering if `continue` should allow `#f` but I'm thinking that the empty string would do the same thing as false would so maybe that's not helpful?

rfindler:
This looks great to me. I can squash and rebase if we're ready.

greghendershott:
On the one hand there might still be some error/omission that will turn up through active dog-fooding. (Which I'm doing; I was using `racket-hash-lang-mode` in Emacs to edit `lang-tools.scrbl`. :smile: Also I plan to merge that to Racket Mode's main branch pretty soon, which possibly means more people trying it.)

On the other hand, we're not close to a release, so it would probably be OK if this is merged now, and another commit ends up tweaking it in (say) a few days/weeks?

rfindler:
Okay, I've squashed and pushed to master [here](https://github.com/racket/drracket/commit/56a5b27f193d1682f742e60102dad3225a0b82bb).

PR #638, opened 2023-09-11
Fix docs for drracket:comment-delimiters

greghendershott:
Strengthen contract.

Fix some region comment examples.

As pointed out by @rfindler on #634.


rfindler:
Thanks!

Issue #641, opened 2023-09-20
Update "DrRacket support for #lang-based Languages" documentation?


greghendershott:
Opening a separate issue about something discussed on #634 but which is more general; prompted by today's https://github.com/greghendershott/racket-mode/issues/668.

- IIUC, ever since the addition of [expeditor](https://docs.racket-lang.org/expeditor/index.html), it's better if a lang can assume just a [color-textoid%](https://docs.racket-lang.org/syntax-color/color-textoid.html) interface, not a full `text%`, for things like [drracket:indentation](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28elem._drracket~3aindentation%29), [drracket:range-indentation](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28elem._drracket~3arange-indentation%29), and [drracket:grouping-position](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28elem._drracket~3agrouping-position%29). That way, it will work with expeditor, as well as tools like the back end support for `racket-hash-lang-mode`.

- However [DrRacket support for #lang-based Languages](https://docs.racket-lang.org/tools/lang-languages-customization.html) still talks about `text%` or `racket:text%`.

  - On the one hand, its advice literally satisfies its title -- making langs work in DrRacket :smile:. And some of the items are in fact only intended to work in DrRacket, like [drracket:toolbar-buttons](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28elem._drracket~3atoolbar-buttons%29) or [definitions-text-surrogate](https://docs.racket-lang.org/tools/lang-languages-customization.html#%28elem._definitions-text-surrogate%29). (Some of these may indeed need a full `text%`.)

  - On the other hand, the advice prevents other items as mentioned above from working in expeditor or other tools.

---

So it seems like the page needs to be updated, or split, or something?

Although I was tempted to make a PR, I wanted to ask first and see what ideas people had?  How to title, organize, etc.?  I'm also fine deferring if someone else like @rfindler prefers to draft it instead.

rfindler:
I think it would be great to use `color-textoid<%>` in any place where that would work, in preference to `text%`! I'm not sure which ones make sense or how hard it would be to get everything lined up properly (maybe just a matter of adjusting the docs and maybe some checks for some other functionality?). I'd be delighted if you're interested in making a PR. I can also try to jump in an help out in places where my experience might make stuff go faster.

greghendershott:
Closed by merge of #642.

PR #642, opened 2023-09-20
Revise lang-tools.scrbl to generalize from DrRacket

greghendershott:
For issue #641.

### DIFF
--- a/drracket/scribblings/tools/lang-tools.scrbl
+++ b/drracket/scribblings/tools/lang-tools.scrbl
@@ -2,15 +2,17 @@
 @(require "common.rkt")
 
 @title[#:tag "lang-languages-customization"]{
- DrRacket support for @tt{#lang}-based Languages
+ Tool support for @tt{#lang}-based Languages
 }
 
-The simplest and best way to extend DrRacket with support
-for a new language is to implement the language via @tt{
- #lang} (see
+A variety of tools can use extra information specified by a language.
+These tools include DrRacket, expeditor, and more.
+
+The simplest and best way to extend tools to support a new language is
+to implement the language via @tt{ #lang} (see
 @secref["hash-languages" #:doc '(lib "scribblings/guide/guide.scrbl")]
-for more details). DrRacket will then use @racket[read-language] to
-find code and values that it uses to customize itself to the language.
+for more details). Tools will then use @racket[read-language] to find
+code and values that it uses to customize itself to the language.
 
 If the call to @racket[read-language] raises an error,
 DrRacket logs the error at the @racket[_debug] level to a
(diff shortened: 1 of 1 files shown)


drracket/scribblings/tools/lang-tools.scrbl
```diff
@@ -74,98 +78,125 @@ The recognized token styles (specified implicitly via @method[color:text<%> star
           @item{@indexed-racket['parenthesis]}
           @item{@indexed-racket['error]}
           @item{@indexed-racket['other]}]
-These precise colors for these identifiers are controlled by the preferences dialog in DrRacket.
+
+The precise colors for these identifiers are controlled by the
+preferences dialog in DrRacket, and by other customization mechanisms
+in other tools.
+}
```
rfindler, line 85: The preferences are actually part of the framework/preferences library which has fairly minimal dependencies and, I believe, gives you back r/g/b values. Maybe it makes sense to elaborate a little bit here for other tool authors?

greghendershott: The "The precise colors for these identifiers are controlled by the preferences dialog in Racket" text already existed. I just added ", and by other customization mechanisms in other tools."

For a tool like Emacs I think it makes sense to map tokens to Emacs named "faces" (e.g. "font-lock-string-face"); these faces in turn can be customized by the user or a theme chosen by the user. This way e.g. strings will look the same across various programming major modes.

So I haven't needed to dig into framework/preferences for this. I don't know enough to write the kind of explanation you want. Could you maybe add this yourself in a later commit once the rest of this looks OK and is merged?

rfindler: Ah, that makes sense. Sounds like the way it was is best, actually!

(I've read down beyond this but not quite finished reading the diff. Hopefully soon!)

drracket/scribblings/tools/lang-tools.scrbl
```diff
@@ -74,98 +78,125 @@ The recognized token styles (specified implicitly via @method[color:text<%> star
...
- to indent lines in DrRacket. It is called with the position containing the line to be
- indented. It is expected to return the number of spaces that should appear at the beginning
- of the line or @racket[#f]. If @racket[#f] is returned,
- DrRacket uses the standard s-expression indentation rules.
+
+When a language's @racket[_get-info] procedure responds to
+@racket['drracket:indentation], it is expected to return a function
+with this contract:
+
+@racketblock[(-> (is-a?/c (or/c racket:text<%> color-textoid<%>))
```
greghendershott, line 95: In this commit I changed `racket:text<%>` to `(or/c racket:text<%> color-textoid<%>)` and added a margin-note just below explaining.

Part of me worries, though, that someone will read this quickly, skip the margin note, and just use `racket:text<%>`.

Instead, I'd like the "read-quickly default" to be `color-textoid<%>`. IOW if people are in a rush, I'd them to do something for more compatibility -- expeditor, Racket Mode, and more, as well as DrRacket. (`racket:text<%>` should be just for backward compatibility for old langs. Or maybe a new lang if they have no choice (if that's even a scenario, I'm not sure?)).

So what do you think about just saying `color-textoid<%>` here, no `or/c`, and leave discussion of `racket:text<%>` to the margin-note?  (Do you see what I mean? If I'm not explaining well let me know and I'll try again. :smile:)

greghendershott: I like your "actually" idea, will try some prose like that and push a new commit.

I think it will be OK if updates are lazy not eager -- OK if some old langs still use full `text%`, and get updated if/as/when someone cares? Someday the lang author does. Or maybe some user of it, who wants it to work also in expeditor or emacs, makes a PR. Or no one does, but that's OK cause no one's really using that lang anymore, or at least not for a long time. Meanwhile it still works in DrRacket.

Anyway that's my starting feeling. Mainly I want new langs, or older langs with new tool support added, to get off on the right foot with `color-textoid<%>`.

rfindler:
I've read over the diff and things generally look good to me.

drracket/scribblings/tools/lang-tools.scrbl
rfindler: > I think it will be OK if updates are lazy not eager -- OK if some old langs still use full `text%`, and get updated if/as/when someone cares? Someday the lang author does. Or maybe some user of it, who wants it to work also in expeditor or emacs, makes a PR. Or no one does, but that's OK cause no one's really using that lang anymore, or at least not for a long time. Meanwhile it still works in DrRacket.
> 
> Anyway that's my starting feeling. Mainly I want new langs, or older langs with new tool support added, to get off on the right foot with `color-textoid<%>`.

I'm okay with that. One niggling worry is that `color-textoid<%>` is missing some important method and it might put people off from this path since they'd just keep using `text%`. We would be able to flush out those kinds of problems earlier if we tried to convert stuff. That said, this is not a today problem! 😄

greghendershott:
Sorry it took me an extra day or two, but, I pushed a commit incorporating the discussion of how to present the preference for languages using `color-textoid<%>`.

With that commit, I think this is ready to merge. (But I'm happy to make further edits/changes if you spot anything else.)

rfindler:
This looks good to me. And having the two commits separate seems like a non-bad thing, so I'll go ahead and merge this.

Thanks!

PR #643, opened 2023-09-21
Update menus.scrbl to cmd.exe

spdegabrielle:
replace command.com reference to `cmd.exe` using terminology from Microsoft documentation 
 https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/windows-commands 
nb command.com was last present in Windows ME.


sorawee:
Thanks!

PR #649, opened 2023-11-10
when receiving syntax objects from 'online-check-syntax logger,

rfindler:
discard them unless their source matches the file that originally triggered the expansion

The goal here is to avoid showing annotations from files that are required (and expanded as part of that process).

related to racket/racket#4811


samth:
Should we merge this?

rfindler:
I'm happy with it but was under the impression that you, @samth , wanted to do more testing. Looking back over the issue, however, I see that my impression may not have been accurate! Do you think there's anything more we should do before merging this?

samth:
I think this is good to merge now.

PR #651, opened 2023-11-15
doc: linkify + fix incorrectly closed bracket + typo

sorawee:
This is a follow-up of cb6a7e88308d4f1f69b

### DIFF
--- a/drracket/scribblings/tools/lang-tools.scrbl
+++ b/drracket/scribblings/tools/lang-tools.scrbl
@@ -1,5 +1,6 @@
 #lang scribble/doc
-@(require "common.rkt")
+@(require "common.rkt"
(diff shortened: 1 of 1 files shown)


sorawee:
@rfindler I am not sure about the last bit (where I changed `text%` to `racket:text%`). Commit cb6a7e88308d4f1f69b has `text%` in the contract box, but has `racket:text<%>` in the margin note. Which one is correct?

rfindler:
Either `racket:text%` or `racket:text<%>` is okay, I think. Maybe the best is to say `an object that implements @racket[racket:text<%>]`.

sorawee:
Thanks, fixed. Does it look good to you now?

samth:
Can we merge this?

PR #656, opened 2023-12-17
Make file saving auto-adapt more to a language's extension

mflatt:
For Windows, move a "*.<ext>" pattern to the front when it matches the default extension, because `put-file` uses that first pattern to determine the default file extension.

For example, for a file that starts `#lang rhombus`, the default extension for saving a file turns into ".rhm" instead of ".rkt".


rfindler:
I was confused by the fancy string munging at first but now I see how it works. 

LGTM.

PR #666, opened 2024-03-30
Surround "custom" with brackets in the language selection dropdown

sorawee:
Prior this PR, the language name in the language selection dropdown could read "Determine language from source custom" with some combinations of settings, which doesn't make sense. Either "custom" should be removed entirely,
or it should be parenthesized. Here, I use brackets to make it similar to the language name shown in the REPL pane.


sorawee:
Before:

Screenshot 2024-03-29 at 6 14 25 PM

After:

Screenshot 2024-03-29 at 6 13 28 PM

PR #679, opened 2024-09-07
Use errortrace-key instead of DrRacket's own key.

mikesperber:
This is so we can pick it up in the teaching languages for reporting error locations.

As discussed with Robby at ICFP.


mikesperber:
(The ensuing changes in `htdp-lib` and `deinprogramm` are ready to go.)

rfindler:
Mike and I just talked in person about this -- this may be the best (only?) way to make this change, but it is possible that DrRacket can avoid instantiating the errortrace unit entirely and just use various the handlers from errortrace directly. It is also possible that DrRacket has to instantiate the unit itself for some reason but, if so, it would be good to be sure that the information that's put into the marks is all in the same format as what errortrace does, to make other uses of that information just work.

mfelleisen:
It’s interesting how this change is essentially “just” a “key change” for cont marks.  

(What’s the lesson for aspiring sw devs?)

mikesperber:
> ... but it is possible that DrRacket can avoid instantiating the errortrace unit entirely and just use various the handlers from errortrace directly. It is also possible that DrRacket has to instantiate the unit itself for some reason but, if so, it would be good to be sure that the information that's put into the marks is all in the same format as what errortrace does, to make other uses of that information just work.
We talked again: DrRacket does more than `errortrace-lib`, so we can't quite use it. But I'll indeed need to adjust the format of the continuations marks.

mikesperber:
@rfindler Could you take another look?

rfindler:
The drracket tests seem to pass.

The word "alligned" isn't spelled right. And probably that comment could be clearer.

We probably also need a test case. The test probably should ensure that the format of the marks is the same as errortrace's format by requiring errortrace and calling a function.

Here's some code that passes when run at the command-line with errortrace enabled and fails when it doesn't. How does that look as the start of a test case for drr?

```
#lang racket
(require errortrace/errortrace-lib
         errortrace/errortrace-key)
(define exn
  (with-handlers ([exn:fail? values])
    (define (f x) (unless (zero? x) (+ (f (- x 1)))))
    (set! f f)
    (f 10)))

(define sp (open-output-string))
(parameterize ([current-error-port sp])
  (errortrace-error-display-handler "abc" exn))

(unless (< 10 (length (continuation-mark-set->list (exn-continuation-marks exn)
                                                   errortrace-key)))
  (error 'errortrace-test "failed.1"))

(define lines
  (for/fold ([in-good-region? #f]
             [stacktrace-lines '()]
             #:result stacktrace-lines)
            ([line (in-lines (open-input-string (get-output-string sp)))])
    (cond
      [(regexp-match? #rx"^ *errortrace" line)
       (values #t '())]
      [(regexp-match? #rx"^ *context" line)
       (values #f stacktrace-lines)]
      [else
       (values in-good-region?
               (if in-good-region? (cons line stacktrace-lines) stacktrace-lines))])))
(unless (< 10 (length lines))
  (error 'errortrace-test "failed.2"))
```

PR #684, opened 2024-09-23
Use new custom action in the Resyntax Autofixer

jackfirth:


shhyou:
Can the workflow be triggered only for `racket/drracket` (and not its fork)?

jackfirth:
@shhyou Not sure, I've never tried to run it on a fork.

shhyou:
This cron job in the workflow is triggered on my fork weekly, but I think it should've been turned off for forks.

PR #693, opened 2024-11-07
Update menus.scrbl

spdegabrielle:
added context menu items to documentation

### DIFF
--- a/drracket/scribblings/drracket/menus.scrbl
+++ b/drracket/scribblings/drracket/menus.scrbl
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
    DrRacket to other languages.}
 
 ]
+
+
+@; ----------------------------------------
+
+@section{@onscreen{Context menu}}
+
+The following additional menu items may be available when activating the context menu in an 
+editing panel. 
+
+@itemize[
+
+ @item{@defmenuitem{Tack/Untack Arrow(s)} Tack binding relationship arrows so
+  they do not disappear. Repeat to untack arrows.}
+ 
+ @item{@defmenuitem{Open Defining File} Opens the file with the currently 
+  selected identifier was defined.}
+
+ @item{@defmenuitem{Jump to Next Bound Occurence} Selects the next occurence of the variable.}
+ 
+ @item{@defmenuitem{Jump to Binding Occurence} Selects where the identifier was bound in the text. 
+If imported the name of the imported library is selected.}
+
+ @item{@defmenuitem{Add Require Prefix} Adds prefix-in, prompts for prefix, and renames
+instances of imported identifier to include the prefix.}
+
+ @item{@defmenuitem{Remove Unused Requires} Refactoring tool to remove unused 
+   requires in module.}
+ 
+ @item{@defmenuitem{Rename} Renames all occurences of selected variable.}
+ 
+ @item{@defmenuitem{Collapse S-expression} Caused the current S-expression to 
+  be renderied as (...)}
+
+ @item{@defmenuitem{Collapse S-expression} Expands S-expression displayed as 
+  (...) back to origninal text.}
+  
+ @item{@defmenuitem{Search in Help Desk for ...} Opens browser and triggers 
+  search of documentation for selected term.}
+
+ @item{@defmenuitem{Save Image...} Save currently selected image to file.}
+]
+
+Standard editng functions are also available via the context menu: Undo, 
+Redo, copy, cut, paste, Select All, and Select All.


spdegabrielle:
includes `Save Image...` as this was added in 8.15

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
...
+ @item{@defmenuitem{Collapse S-expression} Expands S-expression displayed as 
+  (...) back to origninal text.}
+  
+ @item{@defmenuitem{Search in Help Desk for ...} Opens browser and triggers 
+  search of documentation for selected term.}
+
+ @item{@defmenuitem{Save Image...} Save currently selected image to file.}
+]
+
+Standard editng functions are also available via the context menu: Undo,
```
rfindler, line 588: Typo here.

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
...
 
 ]
+
+
+@; ----------------------------------------
+
+@section{@onscreen{Context menu}}
+
+The following additional menu items may be available when activating the context menu in an 
+editing panel.
```
rfindler, line 553: We don't use the phrase "editing panel" either. For this one, I think that "either the definitions or interactions window" is probably the right phrase.

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
    DrRacket to other languages.}
 
 ]
+
+
+@; ----------------------------------------
+
+@section{@onscreen{Context menu}}
+
+The following additional menu items may be available when activating the context menu in an
```
rfindler, line 552: I don't think we use the phrase "context menu" elsewhere in the documentation. In the [`racket/gui` docs](https://docs.racket-lang.org/gui/popup-menu_.html), it is called a "popup menu".

I guess a good idea is to look around in the drracket docs and see what term is used and then either change them all or change this one?

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
...
+
+@; ----------------------------------------
+
+@section{@onscreen{Context menu}}
+
+The following additional menu items may be available when activating the context menu in an 
+editing panel. 
+
+@itemize[
+
```
rfindler, line 556: Some of these items appear in some context and some in others. How about breaking them up into lists based on a description of where they might appear?

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
...
+ @item{@defmenuitem{Collapse S-expression} Caused the current S-expression to 
+  be renderied as (...)}
+
+ @item{@defmenuitem{Collapse S-expression} Expands S-expression displayed as 
+  (...) back to origninal text.}
+  
+ @item{@defmenuitem{Search in Help Desk for ...} Opens browser and triggers 
+  search of documentation for selected term.}
+
+ @item{@defmenuitem{Save Image...} Save currently selected image to file.}
```
rfindler, line 585: I think it saves the one you click on, not the selected one.

rfindler (commented):
Thanks for doing this! If you would like me to take it from here, I'm happy to do so. The changes I'd do are the ones mentioned.

spdegabrielle:
> Thanks for doing this! If you would like me to take it from here, I'm happy to do so. The changes I'd do are the ones mentioned.

Go for it!

drracket/scribblings/drracket/menus.scrbl
spdegabrielle: yes you do! 

> 1 Interface Essentials
> 
> The DrRacket window has three parts: a row of buttons at the top, **two editing panels** in the middle, and a status line at the bottom.

image

drracket/scribblings/drracket/menus.scrbl
spdegabrielle: popup is fine. 

I might be tempted to add 'also known as 'context menu' to the racket gui docs - and add an example to the gallery https://docs.racket-lang.org/gui/Widget_Gallery.html

drracket/scribblings/drracket/menus.scrbl
spdegabrielle: Go for it, but I would not worry too much

drracket/scribblings/drracket/menus.scrbl
spdegabrielle: good point

drracket/scribblings/drracket/menus.scrbl
```diff
@@ -543,3 +543,47 @@ brings the corresponding window to the front.
...
+]
+
+Standard editng functions are also available via the context menu: Undo, 
+Redo, copy, cut, paste, Select All, and Select All.
```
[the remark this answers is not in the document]
rfindler: On second thought, I think what you did is better.

rfindler:
Merged as https://github.com/racket/drracket/commit/888f55dc646651726f75f02cb8942c5d5acd4820.

PR #704, opened 2024-12-27
add 3 units of padding around printed picts

mflatt:
A pict's bounding box does not bound the pixels of the pict's rendered form. That kind of "bounding" box works well for composing pictures, but it's confusing when lines and edges are cut off in a DrRacket interaction. A pict's default conversion to bitmap or SVG form, meanwhile, adds 3 units of padding around a pict to help avoid confusion and edge effects, especially when rendering to Scribble documents (where the passing is compensated by an inset return with the bitmap to preserve alignment). Printing in DrRacket does not convert a pict to a bitmap, but instead wraps it directly in a snip.

This commits adds 3 units of padding around the pict within a snip for its printed form. That padding extends the descent, so baseline alignment still works. Snips don't support horizontal inset, at least not in the same way as Scribble (or potentially other consumers of a pict-to-bitmap conversion), but that seems less important in a value-printing context.

With this change, the output of the following program looks nicer and almost certainly is a better reflection on the programmer's intent:

```
(require pict)
(hline 10 0)
(rectangle 10 10)
```


mflatt:
It would make sense to make the amount of padding configurable in DrRacket. I didn't implement that here, because I think it would rarely get changed. The amount of space for a pict's bitmap conversion is controlled by a parameter, but I have never considered changing it in any application.

PR #708, opened 2025-01-04
`make-visible`: use `port-name-matches?` for symbols

shhyou:
When given symbols, they could be names for unsaved buffers. Directly match the definition window and the interaction window instead. Otherwise, fallback to `find-matching-tab` for `path-strings?` like before.

Related to racket/drracket@d27a4804ab8c252afb06ce94ef77057bdac1d661.

### DIFF
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -3419,33 +3419,31 @@
                     (path->string (normal-case-path (normalize-path p2))))))
 
       (define/override (make-visible filename #:start-pos [start-pos #f] #:end-pos [end-pos start-pos])
-        (let ([tab (find-matching-tab filename)])
-          (when tab
-            (change-to-tab tab)
-            (when (and start-pos end-pos)
-              (define (set-the-position ed)
-                (when (send ed port-name-matches? filename)
-                  (send (send ed get-canvas) focus)
-                  (send ed set-caret-owner #f)
-                  (send ed set-position start-pos end-pos)))
-              (set-the-position (send tab get-defs))
-              (set-the-position (send tab get-ints))))))
+        (match (find-matching-tab/which-editor filename)
+          [(cons tab ed)
+           (change-to-tab tab)
+           (when (and start-pos end-pos)
+             (send (send ed get-canvas) focus)
+             (send ed set-caret-owner #f)
+             (send ed set-position start-pos end-pos))]
+          [#f (void)]))
       
       (define/public (find-matching-tab filename)
-        (define fn-path (if (string? filename)
-                            (string->path filename)
-                            filename))
-        (for/or ([tab (in-list tabs)])
-          (define tab-filename (send (send tab get-defs) get-filename))
-          (and tab-filename
-               (pathname-equal? fn-path tab-filename)
-               tab)))
-      
+        (match (find-matching-tab/which-editor filename)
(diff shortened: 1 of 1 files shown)


rfindler:
I'm not sure if I'm missing something here, but I think that `port-name-matches?` is supposed to accept the path string and work with that (and that'll work when the file has the saved name)?

shhyou:
I don't know. The new code is simply to guarantee that the existing code path does not change because I don't understand the code.

rfindler:
I did only some basic testing, but maybe something like this?

```
$  git diff | cat
diff --git a/drracket/drracket/private/unit.rkt b/drracket/drracket/private/unit.rkt
index e25fa757..334103af 100644
--- a/drracket/drracket/private/unit.rkt
+++ b/drracket/drracket/private/unit.rkt
@@ -3419,33 +3419,29 @@
                     (path->string (normal-case-path (normalize-path p2))))))
 
       (define/override (make-visible filename #:start-pos [start-pos #f] #:end-pos [end-pos start-pos])
-        (let ([tab (find-matching-tab filename)])
-          (when tab
-            (change-to-tab tab)
-            (when (and start-pos end-pos)
-              (define (set-the-position ed)
-                (when (send ed port-name-matches? filename)
-                  (send (send ed get-canvas) focus)
-                  (send ed set-caret-owner #f)
-                  (send ed set-position start-pos end-pos)))
-              (set-the-position (send tab get-defs))
-              (set-the-position (send tab get-ints))))))
+        (match (find-matching-tab/which-editor filename)
+          [(cons tab ed)
+           (change-to-tab tab)
+           (when (and start-pos end-pos)
+             (send (send ed get-canvas) focus)
+             (send ed set-caret-owner #f)
+             (send ed set-position start-pos end-pos))]
+          [#f (void)]))
       
       (define/public (find-matching-tab filename)
-        (define fn-path (if (string? filename)
-                            (string->path filename)
-                            filename))
+        (match-define (cons tab ed) (find-matching-tab/which-editor filename))
+        tab)
+
+      (define/private (find-matching-tab/which-editor filename)
         (for/or ([tab (in-list tabs)])
-          (define tab-filename (send (send tab get-defs) get-filename))
-          (and tab-filename
-               (pathname-equal? fn-path tab-filename)
-               tab)))
+          (define (try ed)
+            (and (send ed port-name-matches? filename)
+                 (cons tab ed)))
+          (or (try (send tab get-defs))
+              (try (send tab get-ints)))))
       
       (define/override (editing-this-file? filename)
-        (ormap (λ (tab)
-                 (or (send (send tab get-defs) port-name-matches? filename)
-                     (send (send tab get-ints) port-name-matches? filename)))
-               tabs))
+        (and (find-matching-tab/which-editor filename) #t))
 
       (define/override (get-all-open-files)
         (filter

```

I can push that to your branch, I think, if you want to try it? What do you think?

shhyou:
Sure! Please feel free to apply any changes. I don't have any idea about the appropriate code organization.

shhyou:
Does `port-name-matches?` convert strings to paths before matching the given name?

shhyou:
This is ready, I think?

rfindler:
> Does `port-name-matches?` convert strings to paths before matching the given name?

It doesn't! And I see there was some conversion code in there before my change. I'll bring that back and merge it.

PR #711, opened 2025-01-09
Align keybindings doc with DrR UI

spdegabrielle:
Change the keybindings page to use the platform independent notation as is present in the 'Show Active Keybindings' keybindings window.


spdegabrielle:
a change to the docs at https://docs.racket-lang.org/drracket/Keyboard_Shortcuts.html

so it matches 
image

rfindler:
LGTM!

PR #753, opened 2025-07-01
add support for reverse-order cycling in `drracket:indentation` and `drracket:range-indentation`

mflatt:
For a language that can offer indentation alternatives, such as Rhombus, allow a reserve-order request via a new argument to a function provided for `drracket:indentation` and `drracket:range-indentation`.

The new argument is a list of options, just in case additional options are needed in the future. Currently, the list is always `'()` or `'(reverse)`.


mflatt:
Revised to add `drracket:range-indentation/reverse-order`, instead.

PR #770, opened 2025-09-26
make check syntax investigate more identifiers

rfindler:
Specifically identifiers that it finds inside origin fields are treated as if they themselves were in the original program, so their origin fields (etc) are all checked.

This may produce some arrows that don't make sense, possibly. I'm not sure.


mflatt:
I've confirmed that this works for a Rhombus improvement so that an arrow goes from the `Posn` use to its binding in

```
#lang rhombus
class Posn(x, y)
#false :: Posn.of(Int, Int)
```

rfindler:
Yay! Thanks.

PR #805, opened 2026-04-08
in the module language, drracket asks the #lang language what the default value of the annotations should be

rfindler:
This is designed to allow the teaching languages to turn on test coverage by default


samth:
This seems to have broken the following test: https://drdr.racket-lang.org/72582/racket/share/pkgs/drracket-test/tests/drracket/online-compilation-zo-creation.rkt (cc @rfindler)

rfindler:
Thanks. I noticed that too but haven't investigated this one yet. Sorry for the delay!

rfindler:
I think I've pushed a fix now.

PR #807, opened 2026-04-28
Update module-lang-test.rkt error messages in accordance with HtDP changes

shhyou:
racket/htdp@8cb765d0 makes the error location available in the messages. Therefore, add them to the tests.

PR #808, opened 2026-04-28
Move some HtDP (test engine) tests to racket/htdp

shhyou:
It'd help to run the test-engine tests with CI whenever Test Engine is updated in racket/htdp, plus there are quite a few HtDP tests in module-lang-tests.rkt. This PR moves them to racket/htdp.


rfindler:
Ah, shoot! I was just fixing the module-lang tests this morning and should have read this PR first! My mistake. I'll adjust.

mfelleisen:
Why are we moving these tests?

shhyou:
@mfelleisen The moved tests are mainly affected by HtDP updates, so I think it would be better if HtDP CI runs them. Right now, DrDr is the only place that runs all of them, and these HtDP tests aren't checked by Github CI in racket/htdp.

shhyou:
@rfindler no worries! I put the first commit in #807 (which is roughly the same as yours) so DrDr can be happy, while the current one (#808) still need some time until the HtDP PR is ready.

mikesperber (commented):
@shhyou What's the motivation for this?

These are essentially integration tests for teaching-languages-on-DrRacket. I'd like to gradually remove the DrRacket dependencies from the HtDP/DeinProgramm packages, but this goes in the opposite direction.

shhyou:
@mikesperber I just opened racket/htdp#264. In summary, I place the integration tests in a standalone package `htdp-drracket-test` that is separated from both `htdp-test` and `drracket-test`. The reason to keep it in racket/htdp is to run the tests for each push to racket/htdp, without having to wait for a full DrDr build.

I consider having full HtDP-DrRacket integration tests useful because there are too many parameters to correctly set up (and I don't know if there will be more in the future). For example, in racket/htdp@e6b79368c I found that the tests didn't check HtDP error messages because of a missing call to `configure`, and in racket/htdp@dc985b3cc I found that the tests were using Racket bindings instead of *SL bindings--the `#%top` binding is still wrong. The tests I added in #689 validate that DrRacket highlights `check-expect` in the Definitions window, and I don't know how to do it without DrRacket. In racket/htdp#229, it also mattered whether the DrRacket buffer was saved or not, which produced different error messages.

I agree that some tests in racket/htdp#264 may not need DrRacket integration. Still, I think running existing HtDP-DrRacket tests for every push to racket/htdp help catch regressions early, which is the goal of racket/htdp#264 and the current PR.

rfindler:
One thought, @shhyou : would it be helpful to have more of the htdp tests into the drracket repo CI? Not sure if that's useful, but I'm certainly open to the idea.

shhyou:
The HtDP (integration) tests here are already in DrRacket CI and get triggered on each push to `racket/drracket`. In contrast, I think there are two advantages for placing HtDP tests in `racket/htdp`:
* Automatically run them for each push to `racket/htdp` and associate the testing results to each commit to the HtDP repository
* Include code updates and tests in a single PR/commit (without having to make two separate commits in two repositories

Also, since the integration tests are in `htdp-drracket-test` which is separated from `htdp-lib` and `htdp-test`, they don't affect normal installation flow.

mikesperber:
Ah, excellent. Thanks for explaining, @shhyou !

mikesperber (approved):
Go ahead!

